PR #3, opened 2016-10-12
make adhesion form overridable by any plugin

guillomovitch:


trasher:
Hello,

Thanks for the PR :)

This affects too many files, it would be better to add tha globally into `Galette\Core\Smarty`.

As I've stated in galette/plugin-fullcard#1; I'd prefer to avoid `global` usage; I'm thinking about something in the plugins class that would permit to override some general entries for example (maybe that should be usefull in the future).

I do not have any precise idea for now, but I'll think about it. I you have suggestions or ideas, feel free to let me know ;)

trasher:
See #4 for an alternative proposal :)

trasher:
#4 have been used instead ;)

PR #4, opened 2016-10-16
Add a preference entry to override pdf adhesion form URl

trasher:


trasher:
From a plugin, we can override the default value using somtehing like `$preferences->pref_adhesion_form_url = 'xyz.php'`; this way we should probably avoid any `global` usage.

The only need is that must be defined at an early stage (on plugin load?) to be sure the value will be the correct one when needed.

PR #5, opened 2016-11-11
Mailmulti pref

boninj:
Add an option allowing use of a same email for multiple accounts.


boninj:
Pull request error: should be in boninj/galette, not in galette/galette
Sorry

trasher:
No problem :) 

Just be aware that such a feature will not be integrated into Galette, as it has already been stated on the mailing list. 

For different members to have the same e-mail address, you should use the members attachment feature.

PR #7, opened 2017-03-31
Develop

guillomovitch:
Dynamic form for advanced search.


galette/templates/default/advanced_search.tpl
```diff
@@ -165,28 +165,68 @@
                 <ul id="fs_sortable" class="fields_list connectedSortable">
 {foreach from=$filters->free_search item=fs}
                     <li>
-                        <select name="free_logical_operator[]">
+                        <select id="logical_operator_selector" name="free_logical_operator[]">
```
trasher: Since we're in a foreach; you may not use `@id` here because it must be unique on the page.

guillomovitch:
Indeed. I just pushed another commit fixing the issue.

trasher:
Manually merged after some changes. Thanks!

PR #11, opened 2017-06-14
Develop+access control

guillomovitch:
New clean PR for fields access control

### DIFF
--- a/galette/lib/Galette/Core/Authentication.php
+++ b/galette/lib/Galette/Core/Authentication.php
@@ -52,6 +52,12 @@ namespace Galette\Core;
 
 abstract class Authentication
 {
+    const ACCESS_USER = 0;
+    const ACCESS_MANAGER = 1;
+    const ACCESS_STAFF = 2;
+    const ACCESS_ADMIN = 3;
+    const ACCESS_SUPERADMIN = 4;
+
     private $login;
     private $name;
     private $surname;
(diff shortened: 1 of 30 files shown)


galette/lib/Galette/Core/Authentication.php
```diff
@@ -292,4 +298,26 @@ public function __set($name, $value)
...
+
+        if ($this->isSuperAdmin()) {
+            return self::ACCESS_SUPERADMIN;
+        } elseif ($this->isAdmin()) {
+            return self::ACCESS_ADMIN;
+        } elseif ($this->isStaff()) {
+            return self::ACCESS_STAFF;
+        } elseif ($this->isGroupManager()) {
+            return self::ACCESS_MANAGER;
+        } else {
```
trasher: Not logged user should probably be added as well, since self-subscription may use it, no?

galette/templates/default/config_fields.tpl
```diff
@@ -16,7 +16,7 @@
                 <li class="listing ">
                     <span class="label">{_T string="Field name"}</span>
                     <span class="yesno">{_T string="Required"}</span>
-                    <span class="yesnoadmin">{_T string="Visible"}</span>
+                    <span class="access">{_T string="Accessibility"}</span>
```
trasher: "Accessibility" does not sounds right to me (same for other occurrences)

guillomovitch:
Current patchset tries to keep the previous semantics by reusing numerical values:
- fields with "admin" visibility are now accessible read/write for admin-only, excluding staff members
- fields with "all" visibility are now accessible read/write for everybody

The same strategy could also get achieved by rewriting corresponding numerical values, which would requires changing database content while upgrading, but would provides additional benefits:
- reordering those values logically (0: most restrictive, 6: less restrictive)
- use a single set of symbolic values, whereas we currently have two different sets, one for core fields, another for dynamic fields

About the property name, current code base use either "visibility" for core field, "perm" for dynamic fields. "Visibility" seems too limited when considering "read only" vs "read/write" difference, whereas "accessibility" seemed fine for me. What about "authorization" or "permission" then ?

galette/lib/Galette/Entity/FieldsConfig.php
```diff
@@ -174,9 +178,6 @@ public function load()
 
             $this->categorized_fields = null;
             foreach ($results as $k) {
-                if ($k->field_id === 'id_adh' && (!isset($preferences) || !$preferences->pref_show_id)) {
-                    $k->visible = self::HIDDEN;
-                }
                 if ($k->field_id === 'parent_id') {
                     $k->visible = self::HIDDEN;
```
trasher: `HIDDEN` no longer existss

PR #19, opened 2018-01-30
Develop

guillomovitch:


guillomovitch:
Better handling of dynamic fields list definitions, independently of any adherent object.

PR #20, opened 2018-02-24
Develop

guillomovitch:
This PR fix the issue of dynamic field descriptors being subject to access control, making the content of an empty adhesion form depend of logged user privileges: only access to dynamic field values should be restricted. It also allows to pass an undefined member to adhesion form class, instead of forcing to create a empty member object when creating an empty adhesion form.

The PR also brings more clarity in dynamic field classes naming:
- DynamicField is a single dynamic field descriptor
- DynamicFieldsSet is a list of dynamic field descriptors
- DynamicFieldsHandle is an aggregation of dynamic field descriptors and values


trasher:
Tests are OK now, thank you :)

There is still an issue on CS (see travis builds reports) to fix, and this will be merged.

guillomovitch:
New attempt, after reorganising changes into 3 commits.

There is a last style 'error' I didn't fix, in DynamicFieldsHandle.php, caused by a multi-line conditional expression which seems easier to read this way.

trasher:
merged, see #21

PR #24, opened 2018-08-26
fix dynamic fields in PDF

boninj:
Dosen't work with empty PDF yet...

### DIFF
--- a/galette/lib/Galette/IO/PdfAdhesionForm.php
+++ b/galette/lib/Galette/IO/PdfAdhesionForm.php
@@ -199,31 +199,31 @@ class PdfAdhesionForm
                 $field_value = '';
                 if ($adh !== null) {
                     $field_values = $adh->getDynamicFields()->getValues($field_id);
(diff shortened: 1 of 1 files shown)


trasher:
Thank you, this has been manually merged in the develop branch:
https://github.com/galette/galette/commit/44d206d706f7df96b80b8229619a58286e50c421

> Dosen't work with empty PDF yet...

I do not understand. Your changes are on dynamic fields values, and there is no dynamic fields values for empty cards.

boninj:
After thinking of it, I'm not sure it's from Galette... It may be errors in my HTML Form code.

In my generated PDF, all the fields are shift: it seems the text fields disappeared and the "special" fields (checkbox, ...) appears instead...
Here's an example: [https://www.cppbauxerre.fr/formulaire_adhesion.pdf](url)

trasher:
OK, thanks for the clarifications. Anyways, proposed fix has been integrated and does fix some issues ;)

PR #33, opened 2019-01-05
Rework extensions and PHP version checks; closes #1225

trasher:
Checks are done from main entry point, each time:
If some extension is removed, it will be caught.

User is redirected to compat page that is displayed
without any translations (because translations requires
intl module).


trasher:
Merged

PR #41, opened 2019-11-21
Remove members selection on mailings

trasher:
As discussed on mailing list:
- remove new mailing button,
- remove useless and unmaintainable members seelction


trasher:
Postponed: this change is not OK, this no longer allows to change recipients list when using an existing mailing as template.
The mailing UI must be rewritten.

PR #42, opened 2019-12-03
Feature/controllers

trasher:


trasher:
Included in https://github.com/galette/galette/pull/43

PR #43, opened 2019-12-06
Dependency injection and routes controllers

trasher:
See:
- https://bugs.galette.eu/issues/1372
- https://bugs.galette.eu/issues/1354
- https://bugs.galette.eu/issues/1355


trasher:
Replaced with #84

PR #47, opened 2020-04-18
Add preference to offer last months of the year

trasher:


trasher:
Need to be unit tested.

PR #52, opened 2020-04-30
Improve groups search

trasher:
Follow and replace #36; without using temporary table. Unit tests has been added.


trasher:
TODO:
* drop sortable, really not needed here,
* clean group data in filters, a simple array is enough

PR #75, opened 2020-10-20
Fix mailing temporary path

trasher:


trasher:
Related issue: https://bugs.galette.eu/issues/1497

PR #111, opened 2021-09-28
Create SECURITY.md

zidingz:
I'd like to report a security issue but cannot find contact instructions on your repository.

If not a hassle, might you kindly add a SECURITY.md file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration!

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security issues to <email>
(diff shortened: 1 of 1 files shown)


trasher (approved):

trasher:
@zidingz have you send details about the issue you found?
Please note that 0.9.5 that has not been officially announced (but which archive exists on the download repository) solves several security issues already; a 0.9.5.1 is on the way and should be released soon.

Of course, if the issue you found is not yet fixed; I'd like it to be part of the incoming release.

Thanks :)

PR #174, opened 2022-06-22
Feature/forms

trasher:
Several changes and fix on member form. Goal is to prepare twig fields components to be used in any form (core or plugin).

I've changed the default class to `inline field`; which seems currently better than only `field`.


trasher:
ping @gagnieray :) I cannot request you as a reviewer until you accept the team invitation to join galette ;)

PR #175, opened 2022-06-23
Refine messages

gagnieray:
Toasts do not reveal to be very convenient for all messages. So I reserved them to success messages or notifications that require a user action. That way, success messages are automatically disposable, and notifications closed when the action is taken.

l also reverted their style to default and moved them to the right bottom corner page so that they are more noticeable that over the page title.

Regarding inline messages, they are restored for warnings and errors with a flash transition and a working dispose icon when javascript is enabled.


gagnieray:
I didn't expect removing images would make the tests fail 😕

Sorry for the repetead "Fix tests" commits...

trasher:
> I didn't expect removing images would make the tests fail confused

My bad, sorry ;)

> Sorry for the repetead "Fix tests" commits...

Oh... I've made **lots** of "Fix tests" commits myself :D Not a problem, I'll rework commits before merge

gagnieray:
> Oh... I've made **lots** of "Fix tests" commits myself :D Not a problem, I'll rework commits before merge

OK thanks !

trasher (approved):

PR #177, opened 2022-06-24
Fix socials networks display on member form

trasher:


trasher:
Diff seems quite huge, but hidding spaces cnages on diff view helps a lot ;)

PR #178, opened 2022-06-24
Gulpfile.js improvements

gagnieray:
Using npm, only 2 commands are now required to build the assets after running composer :

1.  `npm install`
2.  `npm run build`

Once these commands have been run, here is the list of useful commands in the development process : 
* `npm run clean` (removes **every** dependencies, including composer ones)
* `npm run install-deps` (install **all** dependencies, including composer ones, and cleaning first)
* `npm run build` (build the assets without installing the dependencies ; shorter equivalent is `gulp`)
* `npm run rebuild` (all the tasks above at once)
* `npm run watch` (launch the required tasks automatically when modifying the corresponding files, and syncs the changes in the browser when build is done ; shorter equivalent is `gulp watch`)

Gulp tasks can still be run individually if required : 
* `gulp clean` (clean the assets only)
* `gulp theme` (copy Galette's semantic theme in semantic's src folder ; required before using semantic's tasks)
* `gulp styles` (build CSS bundles ; only one for the moment)
* `gulp scripts` (build JS bundles)
* `gulp extras` (copy required extra assets)
* `gulp` (all the tasks above at once and shorter equivalent of `npm run build`)
* `gulp watch` (shorter equivalent of `npm run watch`)

And Semantic/Fomantic's tasks can be lauched from the root of the repository too (their names must not be modified) : 
* `gulp build-css`
* `gulp build-javascript`
* `gulp build-assets`
* `gulp 'build ui'` (all the tasks above at once)


package.json
```diff
@@ -6,13 +6,13 @@
...
   },
+
   "scripts": {
-    "build": "webpack",
-    "build-dist": "webpack --mode=production",
-    "watch": "webpack --watch",
-    "clean": "rm -rf ./webroot/js/libs/",
-    "gulp": "gulp",
-    "gulp-ui": "cp -r ./theme/theme* ./semantic/src && cd semantic && gulp"
+    "clean": "rm -rf node_modules && rm -rf semantic && rm -rf galette/vendor && rm -rf galette/webroot/assets && rm -rf galette/webroot/themes",
```
trasher, line 11: Won't that remove local user css files and js?

trasher (approved):

package.json
gagnieray: Yes indeed. I considered that users who use local css and js do not deploy Galette with git :confused: 

Why not moving those files outside themes ?

In case of git deployments, I can add a gulp task to copy those files from the `ui` folder to a dedicated folder in `webroot`, include the paths in .gitignore, and never delete them in gulp's tasks.

trasher: I think moving them would be a good solution. User specific files should not be stored along with core files. You're almost right asserting users that use those king of files are not using git, but I think we cannot presume they won't ;)

I did not had in mind a gulp copy solution, it should indeed solve that. I should have placed those kind of files in a dedicated folder at the very first, but css/js files were were not built at those times.

gagnieray: All right ! I'll do that then :+1: 

Otherwise, I think that git users might prefer to build a Fomantic theme :smile: 

Actually, with the use of Fomantic, users's customizations with local css files will be somehow quite difficult to achieve cleanly unless they remain very basic.

For proper and clean customizations, creating a Fomantic theme will certainly be the recommended way.
We'll have time to talk about that in an other PR :wink:

trasher: Honestly... The CSS customization is not widely used, this was some kind of difficult for most of Galette users. 
The point was just we should not remove user data - or more probalby ours for tests ;)

Creating themes is absolutely not a priority, there are still a few things remaining before 1.0.0 can be released (and I also have to deal  with my server issues now :/).

gagnieray: > The point was just we should not remove user data - or more probalby ours for tests ;)

Can't the data for the tests be stored in the tests folder ? That's what I already did for the test.gif file in 2293005.

The only other file from `webroot` used in the tests that I know is `galette.png`. Did I miss others ?

> Creating themes is absolutely not a priority, there are still a few things remaining before 1.0.0 can be released

I'm not working on that feature specifically. Creating themes, the way I see it, will simply require to build a fomantic "site" based on galette's theme. That's already possible de facto :slightly_smiling_face: 

FYI, the only 2 things missing, I think, it's the selector in the preferences, and using the "splitted" stylesheets and scripts built by Fomantic. For the moment all these files are copied in the default theme folder, but only the full packaged versions are used in the templates.

Ultimately, using the splitted files would allow to load each one on the required pages only (when/if possible one day). But for now, with gulp, it seems quite easy to separate only the assets of the theme, from the ones of all the components used which will then be moved back in `webroot/assets`.

If it's that easy, I will not go further :wink:

PR #183, opened 2022-06-30
Add aliases for main menu selection

trasher:
The goal is to open the main menu correct entry, and to select (or not) the corresponding entry. With current menu, when editing a member, a contribution, a title, any of entitleds, etc. the main menu is completely closed, and no entry is selected.

The new parameter `aliases` sets up aliases that will always open the corresponding main menu.
The new parameter `sub_select` will prevent menu item to be selected in some cases.

### DIFF
--- a/galette/lib/Galette/Core/Galette.php
+++ b/galette/lib/Galette/Core/Galette.php
@@ -178,23 +178,21 @@ class Galette
                         'title' => _T("View, search into and filter member's list"),
                         'route' => [
                             'name' => 'members',
-                            'args' => []
+                            'aliases' => ['editMember']
                         ]
                     ],
                     [
                         'label' => _T("Advanced search"),
                         'title' => _T("Perform advanced search into members list"),
                         'route' => [
-                            'name' => 'advanced-search',
-                            'args' => []
+                            'name' => 'advanced-search'
                         ]
                     ],
                     [
                         'label' => _T("Saved searches"),
                         'title' => _T("Saved searches"),
                         'route' => [
-                            'name' => 'searches',
-                            'args' => []
+                            'name' => 'searches'
                         ]
                     ]
                 ];
@@ -224,7 +222,8 @@ class Galette
                             'title' => _T("View and filter contributions"),
                             'route' => [
                                 'name' => 'contributions',
-                                'args' => ['type' => 'contributions']
+                                'args' => ['type' => 'contributions'],
+                                'aliases' => ['editContribution']
                             ]
                         ],
                         [
@@ -232,7 +231,8 @@ class Galette
                             'title' => _T("View and filter transactions"),
                             'route' => [
                                 'name' => 'contributions',
-                                'args' => ['type' => 'transactions']
+                                'args' => ['type' => 'transactions'],
+                                'aliases' => ['editTransaction']
                             ]
                         ],
                         [
@@ -297,8 +297,7 @@ class Galette
                             'label' => _T("Manage mailings"),
                             'title' => _T("Manage mailings that has been sent"),
                             'route' => [
-                                'name' => 'mailings',
-                                'aliases' => ['mailing']
+                                'name' => 'mailings'
                             ]
                         ],
                         [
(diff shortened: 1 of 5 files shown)


galette/lib/Galette/Core/Galette.php
```diff
@@ -178,23 +178,22 @@ public static function getMenus(bool $public = false): array
                         'title' => _T("View, search into and filter member's list"),
                         'route' => [
                             'name' => 'members',
-                            'args' => []
+                            'aliases' => ['editMember'],
+                            'sub_select' => false
```
trasher, line 182: I cannot choose beetween members list or member add, no one seems relevant. So, no sub menu will be selected.

galette/lib/Galette/Core/Galette.php
```diff
@@ -224,15 +223,19 @@ public static function getMenus(bool $public = false): array
                             'title' => _T("View and filter contributions"),
                             'route' => [
                                 'name' => 'contributions',
-                                'args' => ['type' => 'contributions']
+                                'args' => ['type' => 'contributions'],
+                                'aliases' => ['editContribution'],
+                                'sub_select' => false
```
trasher, line 228: Here it can be a Contribution or a Donation, I cannot choose beetween contributions list any of add links, no one seems relevant. So, no sub menu will be selected.

galette/lib/Galette/Core/Galette.php
```diff
@@ -224,15 +223,19 @@ public static function getMenus(bool $public = false): array
...
                         ],
                         [
                             'label' => _T("List of transactions"),
                             'title' => _T("View and filter transactions"),
                             'route' => [
                                 'name' => 'contributions',
-                                'args' => ['type' => 'transactions']
+                                'args' => ['type' => 'transactions'],
+                                'aliases' => ['editTransaction'],
+                                'sub_select' => false
```
trasher, line 238: Same as member.

galette/lib/Galette/Core/Galette.php
```diff
@@ -312,7 +314,8 @@ public static function getMenus(bool $public = false): array
                             'label' => _T("Imports"),
                             'title' => _T("Import members from CSV files"),
                             'route' => [
-                                'name' => 'import'
+                                'name' => 'import',
+                                'aliases' => ['importModel']
```
trasher, line 318: Editing model will select the model entry in the menu.

galette/lib/Galette/Core/Galette.php
```diff
@@ -379,7 +382,9 @@ public static function getMenus(bool $public = false): array
                                 'title' => _T("Manage statuses"),
                                 'route' => [
                                     'name' => 'entitleds',
-                                    'args' => ['class' => 'status']
+                                    'args' => ['class' => 'status'],
+                                    'aliases' => ['editEntitled'],
+                                    'sub_select' => false
```
trasher, line 387: Entitleds may be Status or Payment Type. `aliases` do not support arguments, so impossible to select any of those.

galette/lib/Galette/Core/Galette.php
gagnieray: Members can be edited from the list. In my opinion, it's relevant enough. Or at least, more relevant than member add, don't you think ?

galette/lib/Galette/Core/Galette.php
gagnieray: Here too, the list is relevant enough in my opinion, for the same reason as members.

galette/lib/Galette/Core/Galette.php
gagnieray: Same comment :smile:

galette/lib/Galette/Core/Galette.php
gagnieray: At least, the section they belong to is opened.

galette/lib/Galette/Core/Galette.php
trasher: Indeed, you're right :)

PR #184, opened 2022-07-01
Disable hover tooltips on touch devices

gagnieray:
Disable on links and menus only.

Because hovering links doesn't exist on touch devices, let's detect if touch is enable on the device to trigger or not popups with js. Otherwise, popups appear when scrolling the menu, or when a link is touched.


trasher (approved):

trasher:
Manually merged

PR #186, opened 2022-07-06
Use tabs on dynamic fields configuration

trasher:


galette/templates/default/pages/configuration_dynamic_fields.html.twig
```diff
@@ -28,6 +28,14 @@
 
 {% block javascripts %}
         <script type="text/javascript">
+            $('.menu.tabbed').tab({
```
trasher, line 31: Both solutions (path+auto or apisettings) have the same behavior:
- current tab is called from ajax
- no other ajax call is done when switching tabs :/

The call to currently loaded tab should not be done, I've not been able to prevent that (I though using `data-loaded=""` should make the work but it does not)

trasher, line 31: Both solutions (path+auto or apisettings) have the same behavior:
- current tab is called from ajax
- no other ajax call is done when switching tabs :/

The call to currently loaded tab should not be done, I've not been able to prevent that (I though using `data-loaded=""` should make the work but it does not)

gagnieray: This is because tabs are already initialized globally in `common.js` on every menu using the `tabbed` class, and because tabs should be initialized on the activating items.

First solution : remove the `tabbed` class on the menu on line 4 and change the selector.
```suggestion
            $('.pointing.menu .item').tab({
```

galette/templates/default/pages/configuration_dynamic_fields.html.twig
```diff
@@ -1,13 +1,13 @@
...
-    <div class="ui three item top attached stackable tabs menu">
+    <div class="ui stackable pointing inverted menu tabbed">
 {% for key, form in all_forms %}
-        <a href="{{ path_for('configureDynamicFields', {'form_name': key}) }}" class="item{% if form_name == key %} active{% endif %}" data-tab="{{ loop.index }}">{{ form }}</a>
+        <a href="{{ path_for('configureDynamicFields', {'form_name': key}) }}" class="item{% if form_name == key %} active{% endif %}" data-tab="{{ key }}">{{ form }}</a>
 {% endfor %}
     </div>
 {% for key, form in all_forms %}
-    <div class="ui bottom attached{% if form_name == key %} active{% endif %} tab segment" data-tab="{{ loop.index }}">
+    <div class="ui bottom attached{% if form_name == key %} active{% endif %} tab segment" data-tab="{{ key }}">
```
gagnieray, line 10:
```suggestion
    <div class="ui{% if form_name == key %} active{% endif %} tab segment" data-tab="{{ key }}">
```
Segments cannot be attached to pointing menus. This is the reason why some tabs are missing a top border.

galette/templates/default/pages/configuration_dynamic_fields.html.twig
```diff
@@ -1,13 +1,13 @@
 {% extends 'page.html.twig' %}
 
 {% block content %}
-    <div class="ui three item top attached stackable tabs menu">
+    <div class="ui stackable pointing inverted menu tabbed">
```
trasher, line 4:
```suggestion
    <div class="ui stackable pointing inverted menu">
```

galette/templates/default/pages/configuration_dynamic_fields.html.twig
trasher: :+1:

PR #189, opened 2022-07-08
Make as much things as possible work without JS

gagnieray:


trasher:
> The main issue this PR doesn't cover is the tabs that don't have a path. The preferences for example... confused

Hum... That may be changed I guess, but this require some extra work that is maybe not stricly required (see next point).

> Search dropdowns using ajax is another issue too. It's not even working when adding a contribution from a member's profile page

Yes... But I'm almost sure this was already the case with the use of selectize.

Maybe should we now consider than Galette requires javascript to work? As far as I know, nowadays, every browser (including "specific" ones) should be able to run some javascript.

What's your opinion?

gagnieray:
> > The main issue this PR doesn't cover is the tabs that don't have a path. The preferences for example... confused
> 
> Hum... That may be changed I guess, but this require some extra work that is maybe not stricly required (see next point).

You're right. However, I'm not so sure in fact that it requires a "real" path. It needs a way to set the active class on the server side.

> > Search dropdowns using ajax is another issue too. It's not even working when adding a contribution from a member's profile page
> 
> Yes... But I'm almost sure this was already the case with the use of selectize.

Indeed !

And actually in the case of membership fees or donations, it is still possible to add new ones from the list using the mass add contribution action.

So, it is only really an issue in the case of transactions.
 
> Maybe should we now consider than Galette requires javascript to work? As far as I know, nowadays, every browser (including "specific" ones) should be able to run some javascript.
> 
> What's your opinion?

Obviously nowadays javascript is everywhere ! It's not even an issue on mobile devices.

Let's make Galette a progressive web application ! :rofl: 

Moreover, other basic main features would require extra work too to make Galette fully usable without js (sortable lists for example), and more advanced ones simply can't work without (charts, plugins, ...).

But, even if it's only a tiny part of the user base, maybe some might only use basic features and do not trust the minified and concatenated js in the releases, and use [NoScript](https://noscript.net) or [LibreJS](https://www.gnu.org/software/librejs).

Maybe then, it might be a good point, if only it's possible, that Galette follows the [GNU's LibreJS recommendations to set its javascript code free](https://www.gnu.org/software/librejs/free-your-javascript.html) ?

Or at least, to provide a way to chose between using or not the minified assets, and include js sources in the releases ?

trasher:
> And actually in the case of membership fees or donations, it is still possible to add new ones from the list using the mass add contribution action.
> 
> So, it is only really an issue in the case of transactions.

Ok, yes. I'll think about that.

> [...]
> Let's make Galette a progressive web application ! rofl

Sure! ;)

About librejs, I've opened a issue on the tracker, I'll answer there:
https://bugs.galette.eu/issues/1642

gagnieray:
> > So, it is only really an issue in the case of transactions.
> 
> Ok, yes. I'll think about that.

Search dropdowns allow a workaround when js is disabled : its input field can be displayed without javascript and used to give the member's ID. Have a look at 4f75567

> About librejs, I've opened a issue on the tracker, I'll answer there: https://bugs.galette.eu/issues/1642

Allright !

trasher (approved):

PR #211, opened 2023-01-09
Use league\event directly, switch to latest

trasher:


trasher:
New v3 of the lib would require a rewrite of all event system in Galette.
I'm not going to work on that

PR #194, opened 2022-07-11
Fix advanced selection modals

gagnieray:

PR #212, opened 2023-01-10
Fix advanced selection modals

trasher:
Replaces #194 with changes proposed at https://github.com/gagnieray/galette/pull/1


trasher:
I guess this was originally stalled as draft because it may require extra work; but at least current PR fixes several points already, so it's OK for merge to me.

gagnieray:
> I guess this was originally stalled as draft because it may require extra work; but at least current PR fixes several points already, so it's OK for merge to me.

That's right. I thought I could resolve the remaining issues in that PR by myself, but I couldn't find the time to finish the work lately 😅

PR #214, opened 2023-01-15
Prevent the use of base64 images; fixes #1659

gagnieray:
There is no option in summernote editor to disable embeded images :(

So this PR only hides the corresponding field with CSS (some display issues were also fixed at the same time).

Drag and drop of images in the editor has been disabled too.
However, a known issue is preventing this option to work properly at the moment : https://github.com/summernote/summernote/issues/3944#issuecomment-1005459721

I also tried to disable pasting images in the editor without success. The recommended way to achieve this is to use a callback, but I didn't manage to make it work : https://github.com/summernote/summernote/issues/2132

### DIFF
--- a/galette/templates/default/elements/scripts.html.twig
+++ b/galette/templates/default/elements/scripts.html.twig
@@ -91,7 +91,13 @@
                     ],
                     styleTags: [
                         'p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'
-                    ]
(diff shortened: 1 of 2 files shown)


trasher:
I've just added a commit that seems to disable image copypaste (I've tested on firefox 108.0.1).

trasher:
Manually merged; thank you :)

PR #216, opened 2023-01-25
Fix missing existing value on search input

gagnieray:
Is there a reason why the value of the search input is set to `null` here ?

Otherwise, without throwing errors when adding a new member, removing this line easily fixes the missing "city" and "place of birth" that are missing when editing a member.


trasher:
I see no reason to have null here; it's probably a mistake or a copy/paste error

PR #218, opened 2023-01-28
Upgrade Slim to latest major version

trasher:


trasher:
A word about this one... Slim 3 is no longer supported from a while, and several of its dependencies are not compatible with recent PHP versions.
I've maintained a few patches, but that's not meant to be a definitive solution.

At the beginning,  I did not proceed the update because some work was mandatory on Galette itself, but this has been achieved now that we use PHP classes for controllers instead of closures; and now the swithc to Twig is effective (no support for Smarty with Slim v4).

Current PR will  made PHP 8.1 the minimal version since 8.0 is security fixes only, until november 2023.

PR #220, opened 2023-02-01
Use 2 dropdowns for batch actions instead of buttons below lists

gagnieray:


gagnieray:
This PR totally removes the original buttons, and now batch actions only work with js. If it's a problem, it will be very easy to add the buttons back :wink: 

Otherwise, I chose 2 different approches regarding downloads from ajax requests.

For downloads without additional settings (csv, cards, labels), I decided to rely on a standard submit. So after choosing the action in a dropdown, a visible green button is displayed next to it. The user must then click the button to download the file.

But if the following solution is preffered, it can also be applied to those downloads.

For downloads requiring additionnal settings (attendancee sheet), I found code on stackoverflow which works pretty well and allows the files to be downloaded in ajax.

I also thought about letting the browser display the pdf in the modal, so that the user can choose to save the file or not, but I changed my mind (however, I think it is still a possibily to explore ; maybe to provide some kind of "preview" functionnality).

galette/templates/default/elements/scripts.html.twig
```diff
@@ -58,6 +58,79 @@
...
+                        }
+                    } else {
+                        window.location = downloadUrl;
+                    }
+
+                    setTimeout(function () {
+                        URL.revokeObjectURL(downloadUrl);
+                    }, 100); // Cleanup
+                }
+            }
```
gagnieray, line 132: This is the code found on stackoverflow.

trasher:
I'll have to try on my side, but I already have a question: "allows the files to be downloaded in ajax": why should files be downladed in ajax? What problem -if any - do you attempt to fix from existing solution?

As far as I remember, in current version, file is downloaded from the browser just by opening its URL; how the browser deals (show or download a PDF file for example) is per user preference.

I'm not really fan of additional JS because it's code to maintain, I'm not at all confident with JS, and it may be difficult to deal with future browsers versions/changes.
I also wonder if we can include it in Galette; since there is no license specified :/

gagnieray:
> I'll have to try on my side, but I already have a question: "allows the files to be downloaded in ajax": why should files be downladed in ajax? What problem -if any - do you attempt to fix from existing solution?

The original "imperfection" I'd like to improve is a UX one :grin: 

I don't find very convenient that batch actions are only available at the bottom of the list. And one group of buttons already leads to display issues when there are many of them, especially on small screens.

So, the way I wanted to solve this was by using dropdowns. And the place to go for them was logically inline, next to the existing mass selection links.

As I'd like the dropdowns to look like the links, I couldn't use standard selects. I had to use fomantic js dropdowns. And moreover, because I also wanted to make everything work with the modals' actions buttons, I saw no other way to use ajax requests.

But, to be honest, I did things as they came. I would not be surprise if I missed something and overcomplicated everything :sweat_smile: 

> As far as I remember, in current version, file is downloaded from the browser just by opening its URL; how the browser deals (show or download a PDF file for example) is per user preference.

Actually, it is precisely to reproduce the "natural" behaviour of the browser with an ajax response that I ended with this solution. 

All you get from the response is a data. At first I could only display it in an iframe (the reason of my comment about a potential "preview" functionality for PDF files). But I wanted to let the browser do as usual in this case.

> I'm not really fan of additional JS because it's code to maintain, I'm not at all confident with JS, and it may be difficult to deal with future browsers versions/changes. 

I know ! 

When you search the web, this solution seems quite common nowadays (thanks to all those trendy js-only interfaces) ; it seems to be "standard" in all "modern" browsers. One downside though : it should not deal with "heavy" files.

I didn't want neither to include a new library (ex: https://github.com/eligrey/FileSaver.js).

> I also wonder if we can include it in Galette; since there is no license specified :/

Who cares about IE and Safari ? :rofl: 

It could be stripped down to the minimum necessary not to be taken as is : 

```
function _downloadBlob(data, xhr) {
    var filename = "";
    var disposition = xhr.getResponseHeader("Content-Disposition");
    if (disposition && (disposition.indexOf("attachment") !== -1) || disposition.indexOf("filename") !== -1) {
        var filenameRegex = /filename[^;=\n]*=(([""]).*?\2|[^;\n]*)/;
        var matches = filenameRegex.exec(disposition);
        if (matches != null && matches[1])
            filename = matches[1].replace(/[""]/g, "");
    }
    var type = xhr.getResponseHeader("Content-Type");
    var blob = new Blob([data], {type: type});
    var url = window.URL.createObjectURL(blob);
    var a = document.createElement('a');
    a.href = url;
    a.download = filename;
    document.body.appendChild(a);
    a.click();                    
    setTimeout(function () {
        window.URL.revokeObjectURL(url);
    }, 100);
    
}
```

gagnieray:
You were right ! It was totally useless :confused:

trasher:
> I don't find very convenient that batch actions are only available at the bottom of the list. And one group of buttons already leads to display issues when there are many of them, especially on small screens.

My question was just about the JS download part, not the whole work done here ;)

I've not yet tested, I'll do as soon as possible! ;)

Thank you!!

gagnieray:
> Thank you!!

You're welcome ! I hope you will like the result :wink: 

I just added a last commit that hopefully fixes remaining issues. It is also related with a previous comment about a missing confirmation message. 

In fact the message in question should appear if there is was an error when uploading the file. I totally removed this part as I think it can not work here. Correct me if I'm wrong :laughing:

trasher:
I've just tested, that sounds far better than it was before :+1: 

It sounds like PDF and CSV batch are broken. On my test instance, I currently only can generate member cards, even if I choose labels or CSV export. I took a (very) quick look at the code, I've no idea what the problem could be.

trasher:
Looking a bit deeper, it seems the sent request ajax is wrong. Trying to generate a CSV export on one adherent gaves me the following ajax request:

galette_ajax_batch

There are 2 issues: `labels` is not correct, and it should not be an array, it just have to be present (it was previousely the button value).

gagnieray:
> Looking a bit deeper, it seems the sent request ajax is wrong. Trying to generate a CSV export on one adherent gaves me the following ajax request:
> 
> galette_ajax_batch

Indeed, it's strange :thinking: 
Everything is working fine for me, selecting one or more member.

Here is the labels request on my side:
Capture d’écran de 2023-02-04 10-01-47

Maybe you need to run `npm run-script rebuild` if not already done ?

trasher:
I run the npm command, but there is still something wrong unfortunately.

If I first choose to generate label, it ends with a `label: true` in the POST request, that's okay. Then, I choose (without reloading the page) to get member cards. It seems to be okay since I do get the member cards, but the request contains:

```
label: true,
card: true
```

If I choose to get CSV exports: first, I do not have CSV export, but member cards; and now the request contains:

```
label: true,
card: true,
csv: true
```

Trying again to request CSV export, and this time, I have an array:

```
label: true,
card: true,
csv: [
  0: true,
  1: true
]
```

And I always have member cards, since this is the first one in `MembersController::handleBatch()`.

gagnieray:
> If I first choose to generate label, it ends with a `label: true` in the POST request, that's okay. Then, I choose (without reloading the page) to get member cards. It seems to be okay since I do get the member cards, but the request contains: [...]

> And I always have member cards, since this is the first one in `MembersController::handleBatch()`.

Got it, thanks ! :v:

trasher:
Great; all seems working like a charm now! Thanks for the contirbution; it'll be merged soon :)

PR #221, opened 2023-02-04
Feature/forms fixes and improvements

gagnieray:
It started with minor improvements on various forms... And ended with a slightly rework of members' pages :grin:

It's not ready to merge yet as it would need a rebase if #220 is approved.


trasher:
I made a quick test, this seems really good; thanks :)

One question anyways... Why change the order of the main parts? I mean, "Galette relative data" is now displayed after social networks and dynamic fields. While I'm not really against this change, that does not suits the actual setup (that does not include all parts unfortunately).

From a logic point of view, I'm not sure "Galette information" are less important than social networks nor dynamic fields. This should be configurable - I agree - but since you did not change anything in database, I guess you handled this change UI side only, and that seems incorrect.

What's your opinion? Please note I did not yet made a real code review, I've just quick tested ;)

gagnieray:
> I made a quick test, this seems really good; thanks :)

Great! :smile: 

> One question anyways... Why change the order of the main parts? I mean, "Galette relative data" is now displayed after social networks and dynamic fields. While I'm not really against this change, that does not suits the actual setup (that does not include all parts unfortunately).

The main parts of the form can be order in the configuration by drag and drop, just like the fields inside each one of them.

But unless I missed something, socials and the dynamic fields fieldsets can not be order the same way on the form.

That's why I decided to always attach them after the "Contact informations".

This seemed obvious to me for socials. But I agree it is not for the dynamic fields.

> From a logic point of view, I'm not sure "Galette information" are less important than social networks nor dynamic fields. This should be configurable - I agree - but since you did not change anything in database, I guess you handled this change UI side only, and that seems incorrect.
>
> What's your opinion? Please note I did not yet made a real code review, I've just quick tested ;)

Actually it IS configurable in the core fields settings :wink:

Otherwise, maybe we could add a setting to let the admin decide to which part attach the dynamic fields (and let socials attached to the contact informations) ?

trasher:
> But unless I missed something, socials and the dynamic fields fieldsets can not be order the same way on the form.
> 
> That's why I decided to always attach them after the "Contact informations".

Indeed, it is currently not possible to set a position in fieldsets for socials networks and dynamic fields. For the latter, this is an old todo I've never worked on, socials are more recent, I must admit I've never think about that :)

Anyways, doing things right would require a database data update. I prefer not to provide database migration for 1.0, upgrading will already be complicated. It would be possible to add missing fields categories programatically; but for now I'd prefer finish the work on Slim4.

> Otherwise, maybe we could add a setting to let the admin decide to which part attach the dynamic fields (and let socials attached to the contact informations) ?

For me, we have currently two possibles solutions:
1- do not change socials and dynamics fieldsets position at all (as in current stable version),
2- keep your proposition as is.

To be honest, I do not really have a preference, appart the fact proposed solution you propose is maybe a bit "hacky".
In both cases, including socials and dynamics in fields categories configuration will require some work,  the "hack" can certainly be removed at this time; so if you're happy with your solution just keep it right now ;)

gagnieray:
> Indeed, it is currently not possible to set a position in fieldsets for socials networks and dynamic fields. For the latter, this is an old todo I've never worked on, socials are more recent, I must admit I've never think about that :)
> 
> Anyways, doing things right would require a database data update. I prefer not to provide database migration for 1.0, upgrading will already be complicated. It would be possible to add missing fields categories programatically; but for now I'd prefer finish the work on Slim4.

All right !

> For me, we have currently two possibles solutions: 
> 1- do not change socials and dynamics fieldsets position at all (as in current stable version), 
> 2- keep your proposition as is.
> 
> To be honest, I do not really have a preference, appart the fact proposed solution you propose is maybe a bit "hacky". In both cases, including socials and dynamics in fields categories configuration will require some work, the "hack" can certainly be removed at this time; so if you're happy with your solution just keep it right now ;)

Before your code review, I didn't take in consideration that dynamic fields could not always be related to contact information.

So, let's just the socials be attached to contact information for the moment. And the dynamic fields always be positioned after all the other parts :smile:

In all cases, it will only be a question of moving one include from a line to an other, and removing one if statement.

trasher:
> So, let's just the socials be attached to contact information for the moment. And the dynamic fields always be positioned after all the other parts smile

:+1: 

> In all cases, it will only be a question of moving one include from a line to an other, and removing one if statement.

:+1: also :)

PR #222, opened 2023-02-05
Feature/npm composer

trasher:


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rm -rf node_modules && rm -rf semantic && rm -rf galette/vendor && rm -rf galette/webroot/assets && rm -rf galette/webroot/themes",
     "fomantic-install": "cd node_modules/fomantic-ui && npx gulp install",
-    "install-deps": "npm run clean && cd galette && composer install -o && cd .. && npm install && npm run fomantic-install",
+    "install-deps": "npm run clean && cd galette && composer install -o --ignore-platform-reqs && cd .. && npm install && npm run fomantic-install",
```
gagnieray, line 13: Did you do this to try fixing scrutinizer ? 

If so, I rather think the problem relies in the build config for scrutinizer where the failing command run for the tests and on which to add this option is at the moment : `cd galette && wget https://getcomposer.org/download/latest-stable/composer.phar && php composer.phar install`

trasher: o, this change is required for me, default PHP version on my distribution is now 8.1 and therefore will fail on current develop branch.

As for scrutinizer failures, I've changed the configuration trying to make things work for #218 (finally, only php > 8.1 will be allowed because of some dependencies requirements... Also 8.0 is already security fixes only, and next Galette release is not for tomorrow ;)

PR #229, opened 2023-02-09
Feature/css cleaning

gagnieray:


galette/templates/default/components/dynamic_fields.html.twig
```diff
@@ -169,7 +169,7 @@
 <script type="text/javascript">
     {% if not masschange %}
     var _addLnk = function(){
-        return $('<a class="ui small compact icon green button" href="#"><i class="plus icon" title="{{ _T("New occurrence") }}"></i> <span class="sr-only">{{ _T("New occurrence") }}"</span></a>');
```
trasher, line 172: Isn't there any differenciation in Fomantic when we want to hide an element (like icons `i`) or when we want not to be hidden for screen readers?
Since I have no experience with screen readers, I do not know if that really changes something, but from a semantic point of view, "hidden" does not mean the same as "screen-reader only".

galette/templates/default/elements/header.html.twig
```diff
@@ -14,7 +14,6 @@
     {% if file_exists(localstylesheet) %}
         <link rel="stylesheet" type="text/css" href="{{ base_path() }}/{{ constant('GALETTE_THEME') }}galette_local.css" />
     {% endif %}
-        <link rel="stylesheet" type="text/css" href="{{ base_path() }}/assets/css/galette_print.css" media="print" />
```
trasher, line 17: Not sure I understand this removal; is that replaced another way in the framework?

galette/templates/default/components/dynamic_fields.html.twig
gagnieray: > Isn't there any differenciation in Fomantic when we want to hide an element (like icons `i`) or when we want not to be hidden for screen readers?

No there isn't.

>Since I have no experience with screen readers, I do not know if that really changes something, but from a semantic point of view, "hidden" does not mean the same as "screen-reader only".

I do not really have experience with screen readers neither. But I know that the class name doesn't change anything.

Moreover, if such technique should effectively be used to handle screen readers only elements, setting `display:none` is not always the best thing to do. In most cases, it is recommended to move the elements off-screen. But not if it concerns elements navigable with the keyboard (links, forms, ...) or unless they are repositioned to be visible on focus for sighted users... Well... Accessibility is a complex topic !

I will set back the differenciation in the CSS using the proper rule. That will not make galette more accessible, but that way it will keep the original intent.

galette/templates/default/elements/header.html.twig
gagnieray: I moved it in a media query in the theme : https://github.com/galette/galette/pull/229/commits/9789fd8ef9607f8f38fd6ec42d85ea4c771f9b60#diff-e3ee00f5809b9740fa5b5966c7c1f2fa0a7c7ece5a007b63b06e2c8bf79f88f8 :wink:

galette/templates/default/components/dynamic_fields.html.twig
trasher: OK, you can let as is oi was just asking :)

trasher (approved):
OK for me; thank you :)

Could you please rebase on current devleop? I've tried to; but there are few conflicts to deal with; I've preferred to abort ;)

PR #236, opened 2023-02-17
Improvements on emails and pdf models configuration

gagnieray:


trasher:
I see several issues on this one:
- summernote is not displayed on "Automatic emails texts edition" page,
- on PDF models edition page, we only can display source code (see screenshot)
- 
image

I do ensure I've ran npm rebuild, and my browser cache has been cleared this time ;)

gagnieray:
> I see several issues on this one:
> 
>     * summernote is not displayed on "Automatic emails texts edition" page,
> 
>     * on PDF models edition page, we only can display source code (see screenshot)
> 
> 
> image
> 
> I do ensure I've ran npm rebuild, and my browser cache has been cleared this time ;)

On these ones, it is totally on purpose :smile: 

As emails texts are one line strings, when displayed in summernote's codeview, it is actually less convenient than a simple textarea : 

Capture d’écran de 2023-02-19 10-44-25

That's why I ended by only setting monospace font on the inputs.

And concerning the wysiwyg mode, I don't think it is a good idea to edit templates code with it. That's why edition is disabled in this mode and it is only used to "preview" the rendering. For this use case, summernote is then used as a code editor.

If you think editing whith the wysiwyg mode will be usefull, it only requires a slight modification in the controller to instantiate summernote from the template :wink:

trasher:
Oh, I did not understand things that way. That's fine then; thanks for the explanations

gagnieray:
In the case of emails' texts, I've never tested, but is it possible to use html tags in the content ?

I could then enable wysiwyg edition mode, and only set the most simple tags in the toolbar (bold, italic, underline and lists).

trasher:
This is not intended to send administrative emails as complex HTML; I'm not sure I want to officially maintain that :p

PR #238, opened 2023-02-17
Slim v4 error renderer

trasher:


trasher:
This not entirely replace Slim v3 erro handler since this new one will not show Twig errors... Because I'm rendering themp with Twig :/
Seems a limitation on v4; I did not find how to solve that.

PR #241, opened 2023-02-18
Use Labels to display galette modes in navigation

gagnieray:


galette/templates/default/elements/modes.html.twig
```diff
@@ -1,11 +1,12 @@
 {% if login.isSuperAdmin() %}
-    <div class="ui small red inverted center aligned segment" title="{{ _T("You are actually logged-in as superadmin. Some functionnalities may not be available since this is *not* a regular member.") }}">
-        <i class="ui user shield icon"></i>
-        <strong>{{ _T("Superadmin") }}</strong>
-    </div>
+    <span class="ui red icon circular label">
+        <i class="tooltip shield icon" title="{{ _T("You are actually logged-in as superadmin. Some functionnalities may not be available since this is *not* a regular member.") }}"></i>
```
trasher, line 3: I do not like title on `i` element only, I think it's more easy to be find on the whole element.

I've tried moving it to the parent element, and move also the `tooltip` class, but the display is not correct that way, I got standard browser tooltip... I don't know how to get that fixed :/

```suggestion
    <span class="ui red icon circular label tooltip" title="{{ _T("You are actually logged-in as superadmin. Some functionnalities may not be available since this is *not* a regular member.") }}">
        <i class="shield icon"></i>
```

trasher:
Also, I've not tested, but "Demo" mode should really be visible, I wonder if it's still with new proposal

galette/templates/default/elements/modes.html.twig
gagnieray: > I do not like title on `i` element only, I think it's more easy to be find on the whole element.

You're right !

> I've tried moving it to the parent element, and move also the `tooltip` class, but the display is not correct that way, I got standard browser tooltip... I don't know how to get that fixed :/

JS is required.

gagnieray:
> Also, I've not tested, but "Demo" mode should really be visible, I wonder if it's still with new proposal

In my opinion, it is :sweat_smile: 
However, it is not that proheminent on mobile as it is displayed in the menu and thus, not visible until you open the menu.
You'll be judge :wink:

It will not be difficult to move back the mode in the public page template and still use visible label. For now, I just wanted to be consistent with the mode position between private an public pages.

public

public-mobile

private

private-mobile

trasher:
OK, thanks for the screenshots. The only place where "demo" is really less visible is on mobile; but this i snot so important; so keep as it; it's good for me :)

trasher (approved):

PR #247, opened 2023-02-20
Fix tooltips

gagnieray:


galette/templates/default/elements/ajax_members.html.twig
```diff
@@ -34,18 +34,18 @@
                     <td class="{{ rclass }} right">{{ member.id }}</td>
                     <td class="{{ rclass }} username_row">
                     {% if member.isCompany() %}
-                        <i class="ui building outline icon tooltip"><span class="displaynone">{{ _T("Company") }}</span></i>
+                        <i class="ui building outline icon tooltip" title="{{ _T("Company") }}"><span class="displaynone">{{ _T("Company") }}</span></i>
```
trasher, line 37: In current stable version, those tooltips are managed from javascript with the presence of a `span/@sr-only` inside a `.tooltip` element; the inside span contents is used as tooltip contents in that case.

The goal was to prevent complex string repetitions, that historically can be changed/fixed in only one place; leading to confusions or issues in trnalsations or display.

Isn't that possible to use a similar system?

trasher: On another hand, I have to admit this makes UI less usable without javascript enabled; but this is no longer a real goal to have a full JS free UI.

gagnieray: > In current stable version, those tooltips are managed from javascript with the presence of a `span/@sr-only` inside a `.tooltip` element; the inside span contents is used as tooltip contents in that case.
> 
> The goal was to prevent complex string repetitions, that historically can be changed/fixed in only one place; leading to confusions or issues in trnalsations or display.

All right! I thought it was done that way primarily for accessibility purposes. Now I understand the issues you're pointing at.

> Isn't that possible to use a similar system?

Yes, it is possible to use existing inline content. But there are drawbacks. For exemple, I'm not sure the popup content can be a child or a sibling of the element that triggers the popup itself.

gagnieray: > On another hand, I have to admit this makes UI less usable without javascript enabled; but this is no longer a real goal to have a full JS free UI.

In fact, most of the ui components exist in a pure CSS version. However the are more limited, and that often requires heavier html modifications than with js only ; using the js version is most of the time closer to a direct drop-in replacement.

gagnieray: > > Isn't that possible to use a similar system?
> 
> Yes, it is possible to use existing inline content. But there are drawbacks. For exemple, I'm not sure the popup content can be a child or a sibling of the element that triggers the popup itself.

Well... It depends on the cases :confused: 

The equivalent with fomantic requires that the inline content of the popup isn't a child of the trigger. Using only a few classes, it will work without more js code if it's the direct sibling of the trigger.

In the precise case you cited (single `i` tags), no problem.

In the case of "bare" links displaying only an icon using an `i` tag, no big deal neither. Except that no `title` attribute shall be set on the `a` tag (which might be an issue for accessibility... I don't know) unless we control with js when to display `title` on mouse over and out.

In the case of links styled as buttons (or real input buttons), it won't work like you would expect/approve because only the icon would trigger the popup, not the entire button :disappointed:

trasher: If that works in 80% of cases, that's OK for me ;)

Duplicating complex translations is a real issue among time, we'll fix if that cause accessibility (or other!) issues.

Thanks for the job, I'll test/merge it ASAP :-)

trasher:
Be careful, I've rebased against develop. I've added a last commit to prevent text duplication for popups; if It's OK for you, I'll merge this one.

gagnieray:
> Be careful, I've rebased against develop. I've added a last commit to prevent text duplication for popups; if It's OK for you, I'll merge this one.

These are cases where links are styled as icon only buttons. So the tooltip will only popup when the cursor is precisely over the icon, not when it is in the area of the button around the icon :thinking:

trasher:
> These are cases where links are styled as icon only buttons. So the tooltip will only popup when the cursor is precisely over the icon, not when it is in the area of the button around the icon thinking

Ohhhh... Indeed; I've reverted my commit :)

PR #250, opened 2023-02-22
Modals templates refine and refactoring

gagnieray:
A bit more style, consistancy and less js code hopefully :grin:


gagnieray:
Last commit closes #1645, but this doesn't mean that all modals work flawlessly :wink:
At least, all modals now fully benefit from Fomantic UI :tada:

trasher:
> Last commit closes #1645, but this doesn't mean that all modals work flawlessly wink

Most of the work is done, so it's OK for me to close the issue. If there are bugs, users will open new issues :p

> At least, all modals now fully benefit from Fomantic UI and have their content properly (re)loaded tada

Great! Thank you very much :+1: :tada: :fire: :rocket:

PR #252, opened 2023-02-25
Use fomantic's emojis on the dashboard

gagnieray:
Using [Fomantic's emojis](https://fomantic-ui.com/elements/emoji.html) instead of icons, we could add a bit more color back on the dashboard :sparkles: 
What do you think ?

Capture d’écran de 2023-02-25 19-10-10

2 sets of emojis can be used, but only [twemoji](https://twemoji.twitter.com/) has a free (as in :beer:) license (CC-BY 4.0).

The set contains thousands of svg files and would bloat the releases. But it is possible to include only the required ones.

I also chose to include emojis for officials plugins. Auto and Maps ; respectively named `oncoming_automobile` and `map`. So it just requires to change their icon name in their controller.

Capture d’écran de 2023-02-25 19-10-34

But it would be easy to include them in the plugin itself the same way. Then, like for community plugins, it would also be necessary to define the svg file as the `background-image` of `em[data-emoji="unique_name"]::before` in the plugin's CSS.

One thing is missing. As the emojis will be hosted locally, attribution is required. Something like "_Emojis of the dashboard licensed under [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/) by [Twemoji](https://twemoji.twitter.com/)_".

It is stated on the homepage of the project that "_We would consider a mention in the HTML/JS source sufficient also_". So I think I will add it as an html comment for logged in users on the dashboard only. Are you OK with that ?

One last thing : I changed the icon of contributions from `cookies` to `receipt`. I initially chose to use the cookie icon because it is the one used on the members' list on the stable branch.

I don't think money icons are best candidates because contributions are not always a question of money, but more a question of value.

I think the receipt icon is a good "generic" visual representation of what contributions can be : you contribute, you can get a receipt for that :laughing:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -107,6 +107,9 @@ var paths = {
     }, {
       src: './node_modules/jquery/dist/jquery.min.js',
       dest: 'js/'
+    }, {
+      src: './node_modules/twemoji-emojis/vendor/svg/*',
+      dest: 'emojis/'
     }
   ]
 };
@@ -216,15 +219,15 @@ function scripts() {
   return merge(main, masschanges, chartjs, sortablejs, summernote, codemirror, codemirrorxml, codemirrorformatting);
 }
 
-function extras() {
-  main = paths.extras.map(function (extra) {
+function movefiles() {
+  extras = paths.extras.map(function (extra) {
     return gulp.src(extra.src)
       .pipe(gulp.dest(paths.assets.public + extra.dest))
       .pipe(browserSync.stream());
     }
   );
 
-  return merge(main);
+  return merge(extras);
 }
 
 function watch() {
@@ -245,8 +248,8 @@ exports.theme = theme;
 exports.clean = clean;
 exports.styles = styles;
 exports.scripts = scripts;
-exports.extras = extras;
+exports.movefiles = movefiles;
 exports.watch = watch;
 
-var build = gulp.series(theme, clean, styles, scripts, extras, 'build ui', galette);
+var build = gulp.series(theme, clean, styles, scripts, movefiles, 'build ui', galette);
 exports.default = build;
--- a/galette/lib/Galette/Core/Galette.php
+++ b/galette/lib/Galette/Core/Galette.php
@@ -215,7 +215,7 @@ class Galette
             if ($login->isAdmin() || $login->isStaff()) {
                 $menus['contributions'] = [
                     'title' => _T('Contributions'),
-                    'icon' => 'cookie',
+                    'icon' => 'receipt',
                     'items' => [
                         [
                             'label' => _T("List of contributions"),
@@ -553,7 +553,7 @@ class Galette
                         'route' => [
                             'name' => 'members'
                         ],
-                        'icon' => 'users'
+                        'icon' => 'card_box'
                     ],
                     [
                         'label' => _T("Groups"),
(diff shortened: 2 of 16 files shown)


gulpfile.js
```diff
@@ -107,6 +107,39 @@ var paths = {
     }, {
       src: './node_modules/jquery/dist/jquery.min.js',
       dest: 'js/'
+    }, {
+      src: './node_modules/twemoji-emojis/vendor/svg/1f5c3.svg',
```
trasher, line 111: Would copying all svg too large? `1f5c3` is not really self explainning ;)

trasher: Ohhhhh... I've just seen `@emoji-map` :/  That seems a bit complex to maintain, no?

gagnieray: > Would copying all svg too large? 

There are 3689 files. 8,6Mo. 
Around 1Mo compressed.

> `1f5c3` is not really self explainning ;)

This certainly has to do with the [Emoji 14.0 spec](https://www.unicode.org/reports/tr51/tr51-21.html) which the project adheres to and allows different sets to be used.
Fomantic's documentation only cites 2 sets of emojis. 

I suppose that any other set respecting the specs can be used in place easily. Ex: [Openmoji](https://openmoji.org/)

I can create an `emojis` named task instead of simply adding them to `extras`.
Or if you judge it acceptable, include the whole set of emojis.

> Ohhhhh... I've just seen `@emoji-map` :/ That seems a bit complex to maintain, no?

Noooo :wink: 

This is just an override of the complete list maintained in fomantic : https://github.com/fomantic/Fomantic-UI/blob/develop/src/themes/default/elements/emoji.variables

Galette's theme is based on default. So overrinding this variable prevents from adding full css rules in galette's theme ; only the ones mapped here :grin:

trasher: > > Would copying all svg too large?
> 
> There are 3689 files. 8,6Mo. Around 1Mo compressed.

OK for me to copy the whole set, so we do not forget anything in further updates.

> > Ohhhhh... I've just seen `@emoji-map` :/ That seems a bit complex to maintain, no?
> 
> Noooo wink
> 
> This is just an override of the complete list maintained in fomantic : https://github.com/fomantic/Fomantic-UI/blob/develop/src/themes/default/elements/emoji.variables
> 
> Galette's theme is based on default. So overrinding this variable prevents from adding full css rules in galette's theme ; only the ones mapped here grin

Ohhhhh... OK; fine for me then ;)

trasher:
> Using [Fomantic's emojis](https://fomantic-ui.com/elements/emoji.html) instead of icons, we could add a bit more color back on the dashboard sparkles What do you think ?

Previous icons were colored, it's far much better :)

> [...]
> I also chose to include emojis for officials plugins. Auto and Maps ; respectively named `oncoming_automobile` and `map`. So it just requires to change their icon name in their controller.

OK for including in core, that's not really a problem :)

> [...]
> It is stated on the homepage of the project that "_We would consider a mention in the HTML/JS source sufficient also_". So I think I will add it as an html comment for logged in users on the dashboard only. Are you OK with that ?

OK for me.

> [...]
> I think the receipt icon is a good "generic" visual representation of what contributions can be : you contribute, you can get a receipt for that laughing

I've choosed the cookie icon for the same reasons (and that was a wink to Galette). But Icons in dashboard can be different; since they're quite explanatory; I'm OK with your proposals.

gagnieray:
> I've choosed the cookie icon for the same reasons (and that was a wink to Galette)

The wink is still in the footer, with the `cookie bite` next to the version :grin:

trasher (approved):

PR #255, opened 2023-02-28
Move middleware so it can be used when needed

trasher:
On contributions, an error occurs when we just open the dropdown; this PR fix that

### DIFF
--- a/galette/includes/main.inc.php
+++ b/galette/includes/main.inc.php
@@ -161,11 +161,6 @@ if (Galette::MODE_MAINT === GALETTE_MODE && !$container->get('login')->isSuperAd
 }
 
 /**
- * Trailing slash middleware
- */
-$app->add(TrailingSlash::class);
-
-/**
  * Change language middleware
  *
  * Require determineRouteBeforeAppMiddleware to be on.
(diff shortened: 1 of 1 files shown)


trasher:
> Also, there are no longer "next results" at the bottom of the dropdown

But maybe was it intended with FUI? ping @gagnieray

gagnieray:
> > Also, there are no longer "next results" at the bottom of the dropdown
> 
> But maybe was it intended with FUI? ping @gagnieray

To be honest, I don't remember.
Are the search dropdown's results paginated ?
I'll have a look :detective:

trasher:
> Are the search dropdown's results paginated ?

Yes; to solve performances issues: loading all members at once in a dropdown was take much resources; and ends with a very huge HTML page (which is not a good thing neither).

In the `choose_adh.js.twig` template, we call first page only (`url: '{{ url_for("contributionMembers", {"page": 1, "search": "{query}"}) }}'`).

I'll merge this one; we can fix that later (search works, that a correct workaround ;)).

gagnieray:
> > Are the search dropdown's results paginated ?
> 
> Yes; to solve performances issues: loading all members at once in a dropdown was take much resources; and ends with a very huge HTML page (which is not a good thing neither).

Of course ! As I have a very few test members in my database, I didn't even think about that.

> In the `choose_adh.js.twig` template, we call first page only (`url: '{{ url_for("contributionMembers", {"page": 1, "search": "{query}"}) }}'`).

Indeed... I should have noticed :sweat_smile: 

So the dropdown content needs to be reloaded.

One quick solution might be to reuse the existing "reset" button. 
In fact FUI has a "clearable" option for dropdowns and thus this button is useless.

I'll look eventually if it is possible to add a dummy item at the end of the results and reload the content clicking on it without closing the dropdown.

> I'll merge this one; we can fix that later (search works, that a correct workaround ;)).

:+1:

PR #256, opened 2023-02-28
adapt date validation to match date pickers date format

klausellus-wallace:
### Problem

- when trying to create / edit a member in locales like "de_DE" & "en_US" the insert failed with the errors 

  > - Wrong date format (Y-m-d) for Birth date!
  > - Wrong date format (Y-m-d) for Creation date!

### Cause
- the member form's date picker generates dates in the format of `d/m/Y`
- in the current implementation the passed dates for creation & birth are tried to be formatted to the current locale's date format`\DateTime::createFromFormat(__("Y-m-d"), $value);` [Adherent.php](https://github.com/galette/galette/blob/ba185008162bb721e3459b18ac78bb2e17119403/galette/lib/Galette/Entity/Adherent.php#L1268)
- for locales like "fr_FR" the localization coincidentally matched the date pickers date format [galette_fr_FR.utf8.po](https://github.com/galette/galette/blob/develop/galette/lang/galette_fr_FR.utf8.po#L1853)
  ```
  msgid "Y-m-d"
  msgstr "d/m/Y"
  ```
- for others (like "de_DE" & "en_US") it didn't and lead to issues during creation/update of members
    ```
    msgid "Y-m-d"
    msgstr "d.m.Y"
    ```

### Fix

- update the validation's non localized date format to match the date format passed by the date picker

### DIFF
--- a/galette/lib/Galette/Entity/Adherent.php
+++ b/galette/lib/Galette/Entity/Adherent.php
@@ -1268,7 +1268,7 @@ class Adherent
                     $d = \DateTime::createFromFormat(__("Y-m-d"), $value);
                     if ($d === false) {
                         //try with non localized date
-                        $d = \DateTime::createFromFormat("Y-m-d", $value);
+                        $d = \DateTime::createFromFormat("d/m/Y", $value);
                         if ($d === false) {
                             throw new \Exception('Incorrect format');
                         }


galette/lib/Galette/Entity/Adherent.php
```diff
@@ -1268,7 +1268,7 @@ public function validate(string $field, $value, array $values): void
                     $d = \DateTime::createFromFormat(__("Y-m-d"), $value);
                     if ($d === false) {
                         //try with non localized date
-                        $d = \DateTime::createFromFormat("Y-m-d", $value);
+                        $d = \DateTime::createFromFormat("d/m/Y", $value);
```
trasher, line 1271: Well, in facts, `d/m/Y` should be the one tried for `__("Y-m-d")` (the localized version).
Maybe there is a gap; but this line should only act as a fallback to default database known format; when the localized version is not known/incorrect/whatever.

I cannot accept the patch, I'll take a look on my side trying to understand what goes wrong.

trasher (changes requested):

galette/lib/Galette/Entity/Adherent.php
trasher: As far as I understand; the problem is the date picker is not properly loaded. Could you precise which Galette version you use?

klausellus-wallace: hey @trasher, thanks for quick feedback on this.

I'm currently playing around with the latest nightly build (so of course: no support here ;))

The date picker is loading flawlessly as far as I can tell.


It's rather the format the date picker generates the dates. They'll always match `dd/mm/YYYY` regardless of the current locale. I suspect this being the culprit here.

trasher: Yes, the date picker is in cause. 

image

We can see here the format we ask for (`yyyy-mm-dd`) is not hte one in the picker (`dd/mm/yyyy`)

gagnieray: Hi! 

The culprit is the date formatter of the calendar component. For the moment, it is not yet translated (nor really adapted to the required input format I think) : https://github.com/galette/galette/blob/ba185008162bb721e3459b18ac78bb2e17119403/galette/templates/default/elements/js/calendar.js.twig#L61

Since FUI 2.9, it is now possible to use custom formats : 
https://fomantic-ui.com/modules/calendar.html#custom-format
https://fomantic-ui.com/modules/calendar.html#/settings

trasher: I found the issue. In `templates/default/elements/js/calendar.js.twig`, rthe `formatter` returns `return day + '/' + month + '/' + year;`. 
I do not know yet how to fix that

trasher:
I've opened an issue; this require a bit of reflexion/work:
https://bugs.galette.eu/issues/1671

trasher:
CLsoed in favor of #257; but thank you very much for trying to solve :+1:

PR #258, opened 2023-03-01
Feature/paginated search dropdown

gagnieray:
Now using FUI's clearable function on search dropdowns.

It was not possible to add an item at the end of the results to load the next ones.
So a trigger icon has been added on the input field next to the reset button when the dropdown is opened.

Capture d’écran de 2023-03-01 11-40-02

However, I don't see how to get the next page number.

I searched in the master branch to see how it is done at the moment. But it seems the next results button doesn't exist in master neither :confused:

### DIFF
--- a/galette/templates/default/elements/js/choose_adh.js.twig
+++ b/galette/templates/default/elements/js/choose_adh.js.twig
@@ -4,7 +4,6 @@
     var _adhselect = function() {
         $('{{ js_chosen_id }}').dropdown({
             match: 'text',
-            placeholder: '{{ _T("Search for name or ID and pick member") }}',
             apiSettings: {
                 url: '{{ url_for("contributionMembers", {"page": 1, "search": "{query}"}) }}',
                 method: 'post',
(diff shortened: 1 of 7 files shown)


trasher:
I guess you're looking for:
https://github.com/galette/galette/blob/master/galette/templates/default/js_chosen_adh.tpl#L42

Note that navigation buttons were not displayed if there were not enough members to display, or if we were reaching the limit (no prev on first page, and so on).

gagnieray:
> I guess you're looking for: https://github.com/galette/galette/blob/master/galette/templates/default/js_chosen_adh.tpl#L42
> 
> Note that navigation buttons were not displayed if there were not enough members to display, or if we were reaching the limit (no prev on first page, and so on).

Indeed, I missed this one. Thank you :+1:

gagnieray:
Prev/Next buttons are back, and hidden as before if there is only one page of results.

But as they are displayed inside the search input with absolute position, it would be more complicated I think to hide them gracefully when needed. Escpecially since we would have to deal with the reset button as well.

That's why, when reaching the limit, the buttons are simply disabled :wink:

trasher:
I'll tests ASAP, but that seems perfect, thank you :)

trasher (approved):

PR #260, opened 2023-03-02
Run Galette from a subdirectory

trasher:
This is the only way I've found to run properly Galette from a subdirectory.

That will change existing URL, because the `webroot` folder disappears from the URL; adding it throws a 404.


trasher:
This does not works on all cases... 

Plus it prevents using URLs that correspond to existing directories (like `plugins/`)... This is a nonsense.

trasher:
Last commits test another possibility, that requires `RewriteBase` to be correctly set in  galette/`.htaccess` file (not the one under `webroot`!).

With that change, I'm able to get Galette installed in `/var/www/html/public/galette.git`, with a default vhost with `/var/www/html` as `DocumentRoot`. To get all of that working, I has to set the `RewriteBase /public/galette.git/galette/webroot`.

But in the same time, I've setup a VHost with `/var/www/html/public/galette.git` as `DocumentRoot`,  and accordingly set `RewriteBase /galette/webroot` but that does not work :/

trasher:
I've modified the `«ebroot/compat_test.php` so its display the `RewriteBase` to set.

gagnieray:
> Note there is an issue with icons on dashboard; but it will be an issue only if current PR gets merged; and I'm not sure since there is a case that still does not works.

I experienced this issue too and resolved it clearing `data/cache`.

trasher:
> I experienced this issue too and resolved it clearing `data/cache`.

Clearing cache does not fix issue on my side :/

In the CSS file, I got `background-image:url(/assets/emojis/1f5c3.svg)` but the correct url is `background-image:url(/public/galette.git/galette/webroot/assets/emojis/1f5c3.svg)`

trasher:
With the latest commit, there is no longer need for a `RewriteBase` directive; all cases works out of the box! :rocket: 

The dashboard images issue is still present, but I'd like to confirm that everything else work as expected first.

A note: during my tests, I had issues making things work because of a lack in my virtualhost configuration. I had `/var/www/html/public/galette.git` as `DocumentRoot` with the following configuration:
```
<VirtualHost *:80>
    ServerName galette.subdir.localhost

    DocumentRoot /var/www/html/public/galette.git

    <Directory /var/www/html/public/galette.git/>
        Options FollowSymlinks
        AllowOverride All
        Require all granted
    </Directory>
</VirtualHost>
```

I have several paths and some do use symlinks, so I need `Options FollowSymlinks`, but I also had to add `AllowOverride All` in per directory config for the `.htaccess` to work properly.
An alternative would have been to set rewrite rule stuff in vhost config. I think I had issues during tests/development because I forget one of those.

gagnieray:
> With the latest commit, there is no longer need for a `RewriteBase` directive; all cases works out of the box! rocket
> 
> The dashboard images issue is still present, but I'd like to confirm that everything else work as expected first.

On my side, it still doesn't work when calling `/galette` or `/galette/webroot` (redirection error). Nor when calling `/galette/index.php` or `/galette/webroot/index.php` (galette's 404 error page with or without css).

It only works when calling `/galette/webroot/login` for example.

Same on a shared hosting (OVH not to mention it) using a release created from the last commit :confused:

trasher:
I was redacting a message on Galette lists to ask for help for others, but I realized I had issues on login/logout... So, I'll explain here :)

I've tested with 4 distinct configurations (virtualhosts), Galette is always located in `/var/www/html/public/galette.git/`:


1- `DocumentRoot /var/www/html/public/galette.git/galette/webroot`
   URL: http://galette.localhost/login

2- `DocumentRoot /var/www/html/public/galette.git`
   URL: http://galette.subdir.localhost/galette/webroot/login

3- `DocumentRoot /var/www/html`
   URL: http://localhost/public/galette.git/galette/webroot/login

4- `DocumentRoot /var/toto` (we don't care for that case)
   Alias `/magalette/ /var/www/html/public/galette.git/galette/webroot/`
   URL: http://galette.alias.localhost/magalette/login
This last case requires a change in `galette/webroot/.htaccess` to set `RewriteBase /magalette/`

The 1st case is the more correct and simple one (and also the only one working with current develop version).
The 4th case is still usable because it's equally secure (nothing else than `webroot` is exposed from webserver); and it does not require a specific subdomain for Galette.

In all cases, everything seems to be OK, but:
- for other cases than `1`, dashboard icons are not displayed,
- for (almost ?) all other cases than 1; the redirection after login or logout will fail with a 404 or a bad redirect.

So, assuming we're not logged in, if we enter manually the correct URL to the login page (see URLs in above cases), all public pages are accessible without any issues for every cases. Once logged in (if redirection fails and we manually fix it) everything (but dashboard icons and logout) is OK; inclugin for example '`/plugins` URL or any URL relative to Galette or plugins.

gagnieray:
In cases `2` and `3`, a 404 error without CSS remains when calling http://galette.subdir.localhost/galette and http://localhost/public/galette.git/galette

In `galette` folder, removing the `.htaccess` file and changing header's location in the `index.php` file to `webroot` makes things work again.

However, such setup is still subject to issues depending on other `.htaccess` files in the parent folder. For example, I dropped galette in 2 existing vhosts. The first one with a drupal instance, and the other one with a galette 0.9 instance.

With drupal, calling the `galette` folder ends with a 403 error.
With galette 0.9, it ends with a redirection loop.

I think these cases are not galette's concern as the solution relies in the parent's `.htaccess` file. Doesn't it ?

Otherwise, I'm not even sure it is the right way to fix that 404 error.

Indeed, without the changes suggested, this has the advantage to make any folder in `galette` unreachable.

Whereas with the changes suggested, those folders are for the most of them reachable (ex: cron, plugins, vendor, ...).

But maybe it is the same existing behavior with galette 0.9 as this `.htaccess` file doesn't exist in the last release ? :thinking:

trasher:
Thanks for your feedbacks, and for the emojis fix ;)

> In cases `2` and `3`, a 404 error without CSS remains when calling http://galette.subdir.localhost/galette and http://localhost/public/galette.git/galette
> 
> In `galette` folder, removing the `.htaccess` file and changing header's location in the `index.php` file to `webroot` makes things work again.

Indeed, I've changed!

> However, such setup is still subject to issues depending on other `.htaccess` files in the parent folder. For example, I dropped galette in 2 existing vhosts. The first one with a drupal instance, and the other one with a galette 0.9 instance.
> 
> With drupal, calling the `galette` folder ends with a 403 error. With galette 0.9, it ends with a redirection loop.
> 
> I think these cases are not galette's concern as the solution relies in the parent's `.htaccess` file. Doesn't it ?

Indeed, that's not possible. This kind of special cases requires specific adaptation in drupal/whatever configuration.

> 
> Otherwise, I'm not even sure it is the right way to fix that 404 error.
> 
> Indeed, without the changes suggested, this has the advantage to make any folder in `galette` unreachable.

Hum... You talk about the rewriterule? It's not active per default, this cannot be a security system :/

> Whereas with the changes suggested, those folders are for the most of them reachable (ex: cron, plugins, vendor, ...).
> 
> But maybe it is the same existing behavior with galette 0.9 as this `.htaccess` file doesn't exist in the last release ? thinking

Yes, this is how current stable version works.

The only really secure way to go is to expose only `webroot` directory (cases 1 and 4). All other are possible, but less secure, that's what the docs say - and honestly, I guess that's enough ;)

trasher:
A pre-release build is available at:
https://download.tuxfamily.org/galette/dev/galette-1.0.0-subdir.3-20230310-cb5fe25a58.tar.bz2

PR #261, opened 2023-03-02
fix pref-postal-address values in preferences pane

klausellus-wallace:
removed blank spaces & additional curly braces that lead to weird behavior when trying to switch postal address sources (preferences or member)

Prior to these changes, the form would pass `pref_postal_adress:  0}` or `pref_postal_adress:  1}` as form data to `/preferences` endpoint


klausellus-wallace:
FYI - In addition to the above, I've also finalized the german translations for `Core` in weblate.

PR #262, opened 2023-03-02
Fix other preferences values in template

gagnieray:
There were other fields in the preferences template concerned by the issue fixed thanks to @klausellus-wallace (#261) :grin:
I searched through the other templates for similar typos without finding others :wink:


trasher:
I was certainly very tired when I migrate preferences page to Twig :D Thanks to both of you!

PR #264, opened 2023-03-10
Fix over-escaped values in templates; fixes #1673

gagnieray:


galette/templates/default/pages/member_show.html.twig
```diff
@@ -222,7 +222,7 @@
                 {% elseif element.field_id == 'ddn_adh' %}
                                 {{ value }} {{ member.getAge() }}
                 {% else %}
-                                {{ value }}
+                                {{ value|raw }}
```
trasher, line 225: I'll need to triple check that; that may (re)introduce XSS injections; displaying "raw" may be too much.

gagnieray: Of course !!!

In this precise case, the value coming from an input field is already escaped on line 158, and on lines 168, 180 and 182 for textareas :wink:
I think all other values are safe.

trasher: You're right, all values seems to be safe. I was kind of "embarassed with `{% set value = propvalue %}` but looking at that part of code... It seems dumb, so I've removed the condition :)

trasher (approved):

PR #266, opened 2023-03-13
Fix members' contributions list access from menu

gagnieray:


trasher:
With those changes, I guess displayed contributions will not be only the ones for logged in user (in case it's an admin for example).

gagnieray:
> With those changes, I guess displayed contributions will not be only the ones for logged in user (in case it's an admin for example).

Oh... Indeed !

I only looked at the working links on a simple member's dashboard and forgot about admins. I should have read the entire `list` function... :sweat_smile: 

Then, `myList` needs to be fixed because simple members can't access their contributions : 
image

gagnieray:
I checked the existing behaviour on the master branch, and this is already what happens : when a member is an admin, the buttons to list its own contributions are not displayed on the dashboard.

And the "My contributions" and "My transactions" links in the main menu lead to a list of contributions that are not only its own.

Should those links in the main menu really be displayed to admins ?
Isn't that for this reason why they are not existing on the dashboard ?
:thinking:

trasher:
> I checked the existing behaviour on the master branch, and this is already what happens : when a member is an admin, the buttons to list its own contributions are not displayed on the dashboard.
> 
> And the "My contributions" and "My transactions" links in the main menu lead to a list of contributions that are not only its own.
> 
> Should those links in the main menu really be displayed to admins ? Isn't that for this reason why they are not existing on the dashboard ? thinking

I guess there is a mistake on dashboard then :)
Only super-admin should not have those links (since this is not a regular member). For all other profiles, either they're admin or not, they should have the ability to see their own contributions.

I'll take a look at this, but I do not think hiding links was what I had in mind.

trasher:
(commit for docs has been pushed on develop branch already!)

PR #267, opened 2023-03-13
Change default member picture

gagnieray:
To avoid any request regarding the gender of the default picture, it can be replaced with a simple placeholder on the pdf membercard only. 

When a member hasn't upload its picture, I think it is not necessary on the other hand to display a default picture on the member's edit form or info page.


trasher:
I understand and agree with the idea, but just a blank image seems kind of sad for the UI, isn't it?

gagnieray:
> I understand and agree with the idea, but just a blank image seems kind of sad for the UI, isn't it?

On the frontend, the default picture isn't used anymore :
Capture d’écran de 2023-03-13 18-47-40
Capture d’écran de 2023-03-13 18-48-08

It is only displayed on the PDF membercard so that the member can paste a real id photo in place : 
Capture d’écran de 2023-03-13 18-48-56

trasher:
OK, I should have checked! Thanks :)

trasher (approved):

PR #268, opened 2023-03-13
Contributions ACLs

trasher:


trasher:
Replace #266

I guess links on dashboards were limited to non admins because it was using `contributions` route and not `myContributions` one. As far as I remember (I did not check), "my" route is more recent, I should have forget to use it on dashboard :)

gagnieray (approved):

PR #295, opened 2023-09-13
Fix over-escaped value in template; fixes #1690

gagnieray:


trasher (approved):

trasher:
Manually merged, thanks :)

PR #296, opened 2023-09-15
Search on contributions dynamic fields (WIP)

trasher:


trasher:
Searching on contrubtion dynamic fields was not working (at least on dates and boolean fields). This requires tests.

PR #298, opened 2023-09-22
Fix mass adding contribution UI issues

gagnieray:


galette/templates/default/pages/members_list.html.twig
```diff
@@ -807,6 +809,25 @@
                 selector: ".modal-form [id$='rangestart'], .modal-form [id$='rangeend']"
             } %}
         }
+        var _formValidationInModal = function(modal){
+            $(modal +' form').form({
+                inline: true,
+                autoCheckRequired: true,
+                prompt: {
+                    empty: '{name} {{ _T("field required")|e("js") }}'
```
trasher, line 817: This is not correct because terms order may change in some languages. In french for example we'd like `Le champ {name} est requis`.
The best would be to use existing english string `Missing required field %field`; but I've not been able to replace the `%field` part using JS in that case :/

I've tried something like:
```
empty: function (name, value) {
  return '{{ _T("Missing required field %field")|e("js") }}'.replace(/%field/, name);
}
```

But this fails in semantic JS somewhere. I've seen we can use `functions`, but on their examples only in specific fields (and also only using value, not field name).

Since error in displayed just under the related field, we could just use `Required field` without naming it in the message. It also means adding a new translations (which I'd prefer to avoid in RC) but there currently no string that suits the need.

PR #306, opened 2023-09-26
Fix batch actions JS for group managers; fixes #1702

gagnieray:


galette/templates/default/elements/list.html.twig
```diff
@@ -134,7 +134,7 @@
         $(function(){
     {% if batch is defined and batch_actions is defined and batch_actions|length > 0 %}
             var _init_page = function(res){
-        {% if login.isAdmin() or login.isStaff() %}
+        {% if login.isGroupManager() and (preferences.pref_bool_groupsmanagers_exports or preferences.pref_bool_groupsmanagers_mailings) or login.isAdmin() or login.isStaff() %}
```
trasher, line 137: I need to take a deeper look on this one. Change is about groups managers export or mailing capacities, but that also can give them access to delete parts of code.

Could you give me details about which restrictions you found? Is it in core or in a plugin?

gagnieray: > I need to take a deeper look on this one. Change is about groups managers export or mailing capacities, but that also can give them access to delete parts of code.

You're right.
This needs more work :sweat_smile: 

> Could you give me details about which restrictions you found? 

At the moment, it is a Q&D attempt intented to make batch actions modals work for groups managers. Only the list of actions is displayed to them, but the selection and actions are not triggered.

> Is it in core or in a plugin?

It is in core and at least in plugin-events (https://github.com/galette/plugin-events/pull/30/files#diff-d22b309525457f875a21d29f8059c13c6b2496ffef8ddece35a21153dbc94174R183).

trasher: Where is the core problem?

Anyway, I have the idea, and I agree. Since it needs more work, I set this one as draft for now.

gagnieray: > Where is the core problem?

On the contributions list.
The batch actions dropdown is displayed but broken for group managers that are allowed in the preferences to export and send emails.

This PR needs to respect more thoroughly each permission set in the preferences depending on the actions involved.

galette/templates/default/elements/list.html.twig
```diff
@@ -134,7 +134,7 @@
         $(function(){
     {% if batch is defined and batch_actions is defined and batch_actions|length > 0 %}
             var _init_page = function(res){
-        {% if login.isAdmin() or login.isStaff() %}
+        {% if login.isAdmin() or login.isStaff() or login.isGroupManager() %}
```
gagnieray, line 137: Well, this seems to be enough to fix batch actions for group managers in core and existing plugins (https://github.com/galette/plugin-events/pull/30 is required for plugin-events).

Please, have a deeper look to check the possible deleted parts of code you were talking about previously, and change the draft status of this PR at your will.

I checked as admin, staff, group manager and member, and ACLs and rights settings appear to be respected in each case everywhere batch actions are available (members, contributions, transactions, cars models, cars, bookings).

Let me know if I missed anything :blush:

trasher: I think https://github.com/galette/galette/pull/306/commits/e028062a43dd45ee8cff314284717a4d4ec1e893 is enough; if you're OK with that, we can merge this one :)

trasher (approved):

galette/templates/default/elements/list.html.twig
gagnieray: I'm OK with that :+1: :rocket:

PR #308, opened 2023-09-28
Show groupmanager status in members list

gagnieray:


galette/templates/default/pages/members_list.html.twig
```diff
@@ -269,8 +269,12 @@
                         <tr>
                             <th><i class="ui user shield red icon"></i></th>
                             <td class="back">{{ _T('Admin') }}</td>
-                            <th><i class="ui id card alternate orange icon"></i></th>
+                            <th><i class="ui id card alternate red icon"></i></th>
```
trasher, line 272: Why "red"? See https://github.com/galette/galette/pull/308/files#diff-6a061e5c740dc94251e5ef336699fd7126a7fd5e40ded2bb7764d048f7f8d95fR339

red for admins, orange for group manager, andn orange (or another color) for groups members seems correct.

```suggestion
                            <th><i class="ui id card alternate orange icon"></i></th>
```

trasher: BTW, this currently does not work... `isStaff()` always return false on members list due to how members are loaded (I guess this is another "old" bug).

trasher (approved):
I made some changes:
- fix displaying staff members in lists,
- changed staff members icons
- fixed group managers column (red in toc, but orange in list)

gagnieray:
>     * changed staff members icons

It should be changed here also : https://github.com/galette/galette/blob/d017d52114acb1c4cdb05efa57c63f8e2896add7/galette/templates/default/elements/edit_entitleds.html.twig#L75

trasher:
Sorry... I've tired to make quick, and finally made bad work.

trasher (approved):

PR #309, opened 2023-09-28
Show groupmanager status in ajax members list and group persons template

gagnieray:
We missed the ajax template... :sweat_smile:

### DIFF
--- a/galette/templates/default/elements/ajax_members.html.twig
+++ b/galette/templates/default/elements/ajax_members.html.twig
@@ -60,8 +60,13 @@
                         </span>
                     {% elseif member.isStaff() %}
                         <span>
-                            <i class="ui id card alternate orange icon tooltip"></i>
+                            <i class="ui user tie orange icon tooltip"></i>
                             <span class="ui special popup">{{ _T("Staff member") }}</span>
+                        </span>
+                    {% elseif member.getManagedGroups() %}
+                        <span>
+                            <i class="ui users cog orange icon tooltip" aria-hidden="true"></i>
+                            <span class="ui special popup">{{ _T('Group manager') }}</span>
                         </span>
                     {% else %}
                         <i class="ui icon"></i>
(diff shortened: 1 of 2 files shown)


gagnieray:
> We missed the ajax template... 😅

And an other one ! :rofl:

I searched for `isStaff()` through all the templates and I think this time it's OK !

gagnieray:
> [...] I think this time it's OK !

No... There is still something wrong... :disappointed:

trasher (approved):

trasher:
Good catch, thanks! :)

gagnieray:
> Good catch, thanks! :)

However, this time again, `isStaff()` always returns false. Even for groupmanagers 😞

https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/templates/default/elements/group_persons.html.twig#L44-L48

trasher:
On my local instance, all seems correct:
image

In the above image, we can see admins, staff members and group managers icons while selecting groups members from the modal.

gagnieray:
This time I am not talking about the list in the modal, but the list in the group's members tab which is loaded differently and misses `priorite_statut`.

Capture d’écran de 2023-09-29 09-14-00

https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/lib/Galette/Repository/Groups.php#L459-L493

trasher:
> This time I am not talking about the list in the modal, but the list in the group's members tab which is loaded differently and misses `priorite_statut`.
> 
> Capture d’écran de 2023-09-29 09-14-00
> 
> https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/lib/Galette/Repository/Groups.php#L459-L493

Oh, OK! I'll fix

trasher:
`Repository\Groups::getManagerUsers()` is no (longer?) used in core but in Auto plugin. Faulty method here was `Entity\Groups::loadPersons()`; both have been fixed, see a4b4d867500a2c01aff2d649dd08b823872ada73

gagnieray:
Thanks!

PR #310, opened 2023-09-28
Required for plugin-event

gagnieray:
- Related to https://bugs.galette.eu/issues/1707
- https://github.com/galette/plugin-events/pull/33 required

Groupmanagers require permission to request `contributionMembers` to select a member on the booking form.

### DIFF
--- a/galette/includes/core_acls.php
+++ b/galette/includes/core_acls.php
@@ -65,6 +65,7 @@ $core_acls = [
     'contributions'                     => 'member',
     'printContribution'                 => 'member',
     'myContributions'                   => 'member',
+    'contributionMembers'               => 'groupmanager',
     '/(.+)?contribution(.+)?/i'         => 'staff',
     '/(.+)?transaction(.+)?/i'          => 'staff',
     // /Contributions rules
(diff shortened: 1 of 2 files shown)


galette/includes/core_acls.php
```diff
@@ -65,6 +65,7 @@
     'contributions'                     => 'member',
     'printContribution'                 => 'member',
     'myContributions'                   => 'member',
+    'contributionMembers'               => 'groupmanager',
```
trasher, line 68: This route seems not called, either in the plugin (or any other plugin) or in core itself. Did I miss something?

gagnieray: In core, in the templates : 
https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/templates/default/elements/js/choose_adh.js.twig#L27
https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/templates/default/elements/js/choose_adh.js.twig#L84
https://github.com/galette/galette/blob/3db69f0f35c6202489c27dc1faabade467f412ba/galette/templates/default/elements/js/choose_adh.js.twig#L134

trasher: Got it, thanks

trasher (approved):

PR #311, opened 2023-09-28
Hide useless links to members

gagnieray:


galette/templates/default/pages/contributions_list.html.twig
```diff
@@ -96,12 +96,14 @@
             <div class="ui compact vertically fitted segment">
             <div class="ui horizontal list">
             <span class="ui primary ribbon label">
-            <a
+            {% if login.isAdmin() or login.isStaff() %}
```
trasher, line 99: `member.canShow()` here represent parent who access their children contributions. 

If I'm not wrong, this link currently allows them to see the list of their own contributions and also their children. Of course, they still can use the "Clear filter" button on search form (but just like everyone else).
Same should apply to transactions.

Of course, when logged in as a member who have no children, link seems useless; is this why you want to hide the link? Maybe solution can be to check `member.hasChildren()` (and make sure that children are currently loaded at this point - it should be).

gagnieray: Of course, I forgot the children...
I'll do as you suggest, thanks :wink:

gagnieray: Unfortunately, it seems the children are not loaded :confused:

trasher: I'll take an eye

gagnieray: Thanks!

trasher (approved):
Seems OK now

gagnieray:
> Seems OK now

Yes it does ! :v:  :rocket:

PR #317, opened 2023-10-03
Fix missing dynamic contribution fields on advanced search; fixes #1693

gagnieray:


trasher:
Seems to duplicate part of #296

trasher:
Also, this doe snot fix related issue since searches does not work :)

PR #329, opened 2023-10-05
Use stackable display on pagination

gagnieray:
In order to prevent overflows on small screens.

### DIFF
--- a/galette/templates/default/elements/ajax_members.html.twig
+++ b/galette/templates/default/elements/ajax_members.html.twig
@@ -12,20 +12,6 @@
                     </th>
                 </tr>
             </thead>
-            <tfoot>
-                <tr>
-                    <td colspan="3" class="center">
-                        <div class="ui basic left aligned fitted segment">
-                            <div class="ui inverted pagination menu">
-                                <div class="header item">
-                                    {{ _T('Pages:') }}
-                                </div>
-                                {{ pagination|raw }}
-                            </div>
-                        </div>
-                    </td>
-                </tr>
-            </tfoot>
             <tbody>
 {% for member in members_list %}
     {% if excluded is not defined or excluded != member.id %}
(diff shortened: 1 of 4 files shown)


galette/templates/default/elements/ajax_members.html.twig
```diff
@@ -80,6 +66,22 @@
                 <tr><td colspan="3" class="emptylist">{{ _T("no member") }}</td></tr>
 {% endfor %}
             </tbody>
+{% if is_paginated %}
```
trasher, line 69: I had in mind `tfoot` element was always after `thead`, but sources I've found so far are not consistent on that point... :)
Anyway, appart of good or not good practices, I'd prefer to stay consistent (all `tfoot` after either `tbody` or `thead`).

I don't think this has any incidence on rendering, am I wrong?

gagnieray: It actually has an incidence on rendering on small screens, when tables are rendered in a single column :confused:

`tfoot` is displayed right under `thead` (as in the code).

In other cases, though, it it very convenient (ex: statuses management) :grin: :+1: 

But in this precise case, in case of long paginations, it is necessary to scroll all the pagination links before accessing the list of results.

Moving `tfoot` down is the most simple solution.

Moving pagination outside the table fails in "activating" the current selected page link in the pagination (JS would need to be modified).

I will look if it can be solve using JS or only CSS.

trasher: OK, there is a reason, so I'm totally OK; don't worry, and don't spend extra time working on this one if it's already OK ;)

If issue you've spotted is fixed, allright then! Maybe just add a comment specifying tfoot position is in that case importan tin the template, and all will be good.

galette/templates/default/elements/ajax_members.html.twig
```diff
@@ -80,6 +66,22 @@
...
+            <tfoot>
```
trasher, line 70:
```suggestion
            {# keep this one after "tbody" element to prevent display issues #}
            <tfoot>
```

PR #335, opened 2023-10-08
Update JS 3rd party dependencies

trasher:


gagnieray:
> @gagnieray do you think such an update is "safe"? so far, it does not seems to cause issues with FUI, bt Id' like another advice :)

I'll have a look by tomorrow

gagnieray:
It seems you simply ran `npm update`.

It's not the proper way to safely update JS dependencies with npm.

It's better to update the direct dependencies of the project first, as they usually take care of their own dependencies.

Here, I used this tool to ease the process (excluding `del` ; see #340) : https://www.npmjs.com/package/npm-check-updates

trasher:
> It seems you simply ran `npm update`.
> 
> It's not the proper way to safely update JS dependencies with npm.

OK, I had no idea :)

> It's better to update the direct dependencies of the project first, as they usually take care of their own dependencies.
> 
> Here, I used this tool to ease the process (excluding `del` ; see #340) : https://www.npmjs.com/package/npm-check-updates

Great, thank you very much!

PR #346, opened 2023-10-23
Generate and store dark CSS file

trasher:


trasher:
This PR is far to be finished, but I already cause some issues:
- some image paths are incorrect
- dark theme would be enabled for everyone.
- it's currently enabled per default and should be on demand only.

trasher:
Also, maybe using the extension for users who need it is also a good solution (and requires 0 changes Galette side).

PR #351, opened 2023-10-27
Add telemetry step at install

trasher:


trasher:
Globally, it's OK, but steps in footer now displays on 2 lines, this is not the best. I wonder if it would be ok to display just icons, and texts as tooltips (I see no other solution)

gagnieray:
> Globally, it's OK, but steps in footer now displays on 2 lines, this is not the best. I wonder if it would be ok to display just icons, and texts as tooltips (I see no other solution)

Or why not displaying steps vertically ?
https://fomantic-ui.com/elements/step.html#vertical

I'll give a try.

galette/webroot/installer.php
```diff
@@ -111,6 +114,18 @@ function initDbConstants($install)
...
+) {
+    //if we have passed database configuration, define required constants
+    initDbConstants($install);
+
+    if ($install->postCheckDb()) {
+        //while before check db, connection is not checked
+        $zdb = new GaletteDb();
+    }
+}
+
```
gagnieray, line 128: This change causes the following error on my side (wether I choose mysql or postgresql as database type on the database step) :

`PHP Fatal error:  Uncaught Error: Undefined constant "Galette\\Core\\TYPE_DB" in /var/www/galette/galette/lib/Galette/Core/Db.php:110\nStack trace:\n#0 /var/www/galette/galette/install/steps/db_checks.php(55): Galette\\Core\\Db->__construct()\n#1 /var/www/galette/galette/webroot/installer.php(384): include_once('...')\n#2 {main}\n  thrown in /var/www/galette/galette/lib/Galette/Core/Db.php on line 110`

I could only complete the install process reverting this change.
I don't know if it is the right thing to do...

trasher: Indeed, that cannot work... We check if we're at database checks steps; but steps are defined after. Just reverting should not solve the issue; I do not remember right now why I've moved this part, but I guess I had an issue with the new step added.

I'll investigate

trasher: Fixed in last commit

gagnieray: > Fixed in last commit

Now an error is thrown on the database installation step :

`PHP Fatal error:  Uncaught ErrorException: Undefined variable $zdb in /var/www/galette/galette/install/steps/db_install.php:41\nStack trace:\n#0 /var/www/galette/galette/install/steps/db_install.php(41): {closure}()\n#1 /var/www/galette/webroot/installer.php(303): include_once('...')\n#2 {main}\n  thrown in /var/www/galette/galette/install/steps/db_install.php on line 41, referer: http://galette.localhost/installer.php`

trasher: Argh... Indeed, I've tried to make things fast, and I just made things wrong :D

trasher: It should be OK now

gagnieray: Yes :+1:

PR #354, opened 2023-10-30
Display installation steps vertically

gagnieray:


gagnieray:
> But, with current PR, I'm always redirected to installation type after I enter my DB information :( As far as I can see; there is a `stepback_btn` present in POST and it should not (since I've not find a fix, it may not be the only issue).

That happens when hitting `Enter` key.
The problem comes from the order of the submit buttons in the HTML code.
This will be easily solved :wink:

trasher:
> That happens when hitting `Enter` key. [...]

Oh! Indeed, I did not pay attention...

trasher:
I've rebased on my branch to include misssed variables/constants; and I've added a commit to restore buttons order  (21d4e8aaf6325470791759730e170b668bdb5d1b).
Everything seems OK on my side.

gagnieray:
On the frontend side, I think it's OK now :smile: 

But I don't know how to properly fix the unit tests...

And I discovered 2 other issues :

1. with postgresql only : database install fails on an existing database starting with `DROP TABLE IF EXISTS galette_paymenttypes .............. KO`

2. with both postgresql and myslql/mariadb : when database check fails, it's impossible to go back to enter the connection parameters again

trasher:
> On the frontend side, I think it's OK now 😄

Great, thank you :)

> But I don't know how to properly fix the unit tests...

I'll take a look; I did not pay attention tests were failing ^^
 
> And I discovered 2 other issues :
> 
>     1. with postgresql only : database install fails on an existing database starting with `DROP TABLE IF EXISTS galette_paymenttypes .............. KO`

Oh :/ I'll check

>     2. with both postgresql and myslql/mariadb : when database check fails, it's impossible to go back to enter the connection parameters again

Hum... This looks like another regression on my original changes; I'll try to fix.

trasher (approved):

trasher:
>     1. with postgresql only : database install fails on an existing database starting with `DROP TABLE IF EXISTS galette_paymenttypes .............. KO`

I've fixed that on #351 :)

PR #355, opened 2023-11-05
Some slight accessibility related improvements

gagnieray:
Mainly modifications of contrasts.

Otherwise, the `aria-hidden` attribute was already defined on some icons but not all. I think it won't hurt to add it to all of them :grin:


tests/Galette/Core/tests/units/CheckModules.php
```diff
@@ -102,7 +102,7 @@ public function testAllKO()
 
         $html = $checks->toHtml();
         $this->assertStringNotContainsString('green check icon', $html);
-        $this->assertSame(1026, strlen($html));
+        $this->assertSame(1197, strlen($html));
```
gagnieray, line 105: 19 new characters (` aria-hidden="true"`) x 9 tests = 171 + 1026 = 1197

trasher (approved):

PR #356, opened 2023-11-06
Try to fix calendar widget for #1739

trasher:
I've tried to fix... Calendar widget seems not set up properly.

Anyways, issue on date selection is "fixed" since we do not have any widget (as it was in 0.9.x).

### DIFF
--- a/galette/templates/default/pages/transaction_form.html.twig
+++ b/galette/templates/default/pages/transaction_form.html.twig
@@ -240,8 +240,8 @@
                 $('.contributions-selection .contribution_row input[type=checkbox]').hide();
 
                 //Initialize Fomantic components
-                $('.contributions-selection .dropdown').dropdown()
-                $('[id$="rangestart"], [id$="rangeend"]').calendar();
+                $('.contributions-selection .dropdown').dropdown();
+                {% include "elements/js/calendar.js.twig" %}
 
                 //Deactivate contributions list links
                 $('.contributions-selection tbody a').click(function(){
--- a/galette/lib/Galette/Controllers/Crud/ContributionsController.php
+++ b/galette/lib/Galette/Controllers/Crud/ContributionsController.php
@@ -401,14 +401,6 @@ class ContributionsController extends CrudController
         $ajax = false;
         $get = $request->getQueryParams();
 
-        if (
-            ($request->getHeaderLine('X-Requested-With') === 'XMLHttpRequest')
-            || isset($get['ajax'])
-            && $get['ajax'] == 'true'
-        ) {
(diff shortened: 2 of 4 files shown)


trasher:
> Anyways, issue on date selection is "fixed" since we do not have any widget (as it was in 0.9.x).

To me, that may be a correct fix. Searching for a contribution from translactions list is certainly not the most used feature of Galette.

galette/templates/default/pages/transaction_form.html.twig
```diff
@@ -240,8 +240,8 @@
                 $('.contributions-selection .contribution_row input[type=checkbox]').hide();
 
                 //Initialize Fomantic components
-                $('.contributions-selection .dropdown').dropdown()
-                $('[id$="rangestart"], [id$="rangeend"]').calendar();
+                $('.contributions-selection .dropdown').dropdown();
+                {% include "elements/js/calendar.js.twig" with {selector: ".contributions-selection #start_date_filter, .contributions-selection #end_date_filter"} %}
```
gagnieray, line 244: Don't change the default selector for the widget to work properly :

```suggestion
                {% include "elements/js/calendar.js.twig" %}
``` 

However, I noticed a strange behaviour (with or without the widget properly loaded) : when filtering by dates, the list in the modal is reloaded with the correct results, but when reopening the modal afterwards (with or without selecting a contribution), the results are different. Im my case, when first filtering in the modal, I get 2 results. And then, reopening the modal, only one result is displayed.

gagnieray (changes requested):

galette/templates/default/pages/transaction_form.html.twig
trasher: I'll test this afternoon for the date picker, thanks.

trasher: OK. From a transaction, I've opened contributions list popup and then filtered on a payment type. I can see here that several contributions are shown but should not:
- already linked to a transaction
- having an amount greater than remaining amount to dispatch in my transaction.

When I close the popup and then reopen it, the payment type filter is still applied; but other restrictions are correctly applied. This second is the correct result as far as I can see.

I also suspect an issue in session recording (like something stored in session but not used immediately); but that's kind of hard to be sure, I'll have to make extra tests.

trasher:
I think my last commit fix all remaining cases.

What I supposed to be a session issue was in fact a double load, but only when using the "Clear filter" button.

gagnieray:
> I think my last commit fix all remaining cases.

Yes! :clap: :rocket: :smile:

PR #361, opened 2023-11-16
Fix tables rows backgrounds colors on mobile

gagnieray:


galette/templates/default/elements/ajax_members.html.twig
```diff
@@ -16,9 +16,9 @@
 {% for member in members_list %}
     {% if excluded is not defined or excluded != member.id %}
         {% set rclass = member.getRowClass() %}
-                <tr>
-                    <td class="{{ rclass }} right">{{ member.id }}</td>
-                    <td class="{{ rclass }} username_row">
+                <tr class="{{ rclass }}">
+                    <td class="right">{{ member.id }}</td>
+                    <td class="username_row">
```
trasher, line 21: There is another `<td>` with a `rclass` that can be removed

galette/templates/default/pages/contributions_list.html.twig
```diff
@@ -228,8 +228,8 @@
             {% set ctype = constant('Galette\\Entity\\Contribution::TYPE_DONATION') %}
         {% endif %}
 
-        <tr{% if mode == 'ajax' %} class="contribution_row" id="row_{{ contribution.id }}"{% endif %}>
-            <td class="{{ cclass }}" data-scope="row">
+        <tr{% if mode == 'ajax' %} class="contribution_row {{ cclass }}" id="row_{{ contribution.id }}"{% else %} class="{{ cclass }}"{% endif %}>
+            <td data-scope="row">
```
trasher, line 232: `cclass` should be removed on other `td`, no?

trasher:
I've rebased on current develop, and added a commit that removes row class also used on cells, and I've also removed variable declaration when it's used only once.

If you're OK with that @gagnieray; I'll merge :)

gagnieray:
> If you're OK with that @gagnieray; I'll merge :)

Yes I am ! :smile:

PR #364, opened 2023-11-19
Some more accessibility improvements

gagnieray:


galette/templates/default/elements/logged_user.html.twig
```diff
@@ -28,7 +28,7 @@
         </div>
     {% else %}
         {% if not login.getCompactMenu() %}
-            <div class="{{ component_classes }}">
+            <nav class="{{ component_classes }}" aria-label="{{ _T('User menu') }}">
```
trasher, line 31: I'm not convinced adding "anything menu" is really required (and also, that adds new strings to translate, I'd prefer to avoid).

galette/templates/default/pages/index.html.twig
```diff
@@ -10,13 +10,13 @@
...
                                 <div class="ui left icon input">
-                                    <i class="user icon" aria-hidden="true"></i><span class="displaynone"><label for="login">{{ _T("Username:") }}</label></span>
+                                    <i class="user icon" aria-hidden="true"></i><label for="login" class="visually-hidden">{{ _T("Username:") }}</label>
                                     <input type="text" name="login" id="login" autofocus placeholder="{{ _T("Username:") }}"/>
                                 </div>
                             </div>
                             <div class="field">
                                 <div class="ui left icon input">
-                                    <i class="lock icon" aria-hidden="true"></i><span class="displaynone"><label for="password">{{ _T("Password:") }}</label></span>
+                                    <i class="lock icon" aria-hidden="true"></i><span class="visually-hidden"><label for="password">{{ _T("Password:") }}</label></span>
```
trasher, line 19:
```suggestion
                                    <i class="lock icon" aria-hidden="true"></i><label for="password" class="visually-hidden">{{ _T("Password:") }}</label>
```

galette/templates/default/elements/logged_user.html.twig
trasher: After some quick search, it seems required, yes. And for "User menu", is seems makes sense.
I still doubt for some labels ("Secondary menu" or "Top menu" for example: I'm not sure myself what in the UI it represents). I've nothing better to propose right now (and I have to leave), and I do not know if it's better to keep a "bad" label, or not add label at all.

galette/templates/default/elements/navigation/navigation_topbar.html.twig
```diff
@@ -1,9 +1,9 @@
...
-    <div class="ui fluid container">
-        <a class="toc item" tabindex="0">
-            <i class="sidebar icon" aria-hidden="true"></i>
-            <span class="displaynone">{{ _T("Menu") }}</span>
-        </a>
+    <a class="toc item" tabindex="0">
+        <i class="sidebar icon" aria-hidden="true"></i>
+        <span class="visually-hidden">{{ _T("Menu") }}</span>
+    </a>
+    <nav class="ui fluid container" aria-label="{{ _T('Top menu') }}">
```
trasher, line 6: I really do not like the label content. Also, semantically speaking, it is a menu on public pages, yes; but on logged in pages, it only is useful on mobile view, and is not in that cas really a menu but only a button to display/hide the menu.
I do not have propositions right now :/

galette/templates/default/page.html.twig
```diff
@@ -23,18 +24,12 @@
...
-                        <div class="ui center aligned segment">
-                            <h1>
-{% endif %}
-                                {% block page_title %}{{ page_title }}{% endblock %}
-                            </h1>
-{% if login.isLogged() %}
-                        </div>
                         <div class="ui right aligned compact segment">
-                            <div class="ui compact tiny menu">
+                            <nav class="ui compact tiny menu" aria-label="{{ _T('Secondary menu') }}">
```
trasher, line 32: This is not a menu on a semantic point of view
```suggestion
                            <nav class="ui compact tiny menu">
```

trasher (changes requested):

galette/templates/default/elements/logged_user.html.twig
trasher: I searched a bit, see my newest comments ;)

galette/templates/default/elements/navigation/navigation_topbar.html.twig
gagnieray: I don't really understand if your are talking about the "Menu" button or the "Top menu" `<nav>` element :thinking: 

**Regarding the "Menu" button**, it is outside the following `<nav>` element (which, indeed, is only displayed to unlogged users). So, it is simply a link to access the sidebar menu, and I think it is OK this way semantically speaking.

Moreover, when using the keyboard, this button is the second element in the tabulation key order, just after the "Skip to content" link. Which makes sense : if the user didn't wish to skip to content, it is logically to access the navigation.

**Regarding the "Top menu" `<nav>` element**, I suppose it could be OK if its `aria-label` is changed to "Main menu", and if I apply on it at the same time a `display: none` when the user is logged. This way in each case (mobile/desktop, un/logged, etc.), only one `nav` element labelled "Main menu" could be read by screen readers, as others are then hidden with `display: none` (this is only suppositions ; I am not a screen reader user nore an accessibility expert :wink:).

Otherwise, the mobile menu would require other improvements regarding accessibility using screen readers and/or the keyboard. For example, the text of the "Menu" button should actually be "Open the menu" by default, and changed to "Close the menu" when the menu is displayed. And I don't think that the fact that the button and the mobile menu are not children elements of the same parent is actually for screen readers.

But I don't want to go further regarding accessibility with screen readers for the moment. It seems to be a huge topic as far as I have read :sweat_smile:

galette/templates/default/page.html.twig
gagnieray: It would if it only contained the language menu and was labeled as such.
But because it can show an existing email "notification" link, I chose to name it that way.

What if I move the existing email link elsewhere ? :thinking:

trasher: OK, I guess I understand... 

Existing emailing can be moved, that's not a main feature (and we also get reminder a mailing already exists when creating a new one from members list).
I guess we can move this as an extra entry in main menu just before/after "Manage mailings".

trasher: (or fource, it also can be moved elsewhere if you have a better idea :D)

galette/templates/default/elements/navigation/navigation_topbar.html.twig
trasher: Thanks for the clarifications :)

I was talking about the "Top menu" label; that does not means anything really (and that also can be moved elsewhere in the future).
I guess we cannot use "Main menu" label once logged because the "real" "Main menu" can be displayed:
image

On the above screenshot, we can see it's a real menu only when user is not logged in; and therefore can be labelled "Main menu" whe no one is logged, and have no label once user has logged in (in facts, menu entries should not be rendered at all once logged in - but that is not a big issue).

For sure we have to limit the changes on accessibility (I was about to propose this one to be the very last before I release a RC3, and I'd like final release before the end of the month).

gagnieray: > I guess we cannot use "Main menu" label once logged because the "real" "Main menu" can be displayed: image
> 
> On the above screenshot, we can see it's a real menu only when user is not logged in; and therefore can be labelled "Main menu" whe no one is logged, and have no label once user has logged in (in facts, menu entries should not be rendered at all once logged in - but that is not a big issue).

I'll have a deeper look to this.

> For sure we have to limit the changes on accessibility (I was about to propose this one to be the very last before I release a RC3, 

For the final release, I think all recent changes were the best that could be done quite "easily", and were necessary or at least could benefit keyboard usability/accessibility :grin: 

> and I'd like final release before the end of the month).

Great ! :smiley:

galette/templates/default/page.html.twig
gagnieray: > (of course, it also can be moved elsewhere if you have a better idea :D)

You'll judge if it's better, but yes I have an other idea :wink:

galette/templates/default/elements/navigation/navigation_topbar.html.twig
gagnieray: > > I guess we cannot use "Main menu" label once logged because the "real" "Main menu" can be displayed: image
> > On the above screenshot, we can see it's a real menu only when user is not logged in; and therefore can be labelled "Main menu" whe no one is logged, and have no label once user has logged in (in facts, menu entries should not be rendered at all once logged in - but that is not a big issue).
> 
> I'll have a deeper look to this.

OK. I ended up with a simple solution in my upcoming changes : ensuring that _only one_ `<nav>` element is read by the screen reader when the user is unlogged or displaying the mobile view. This way, there is no real need (at the moment) to set a label to the involved `<nav>` elements in the templates.

Result with Orca when the user is unlogged in desktop view :
image

Result with Orca when the user is unlogged in mobile view (menu closed) :
image

Result with Orca when the user is unlogged in mobile view (menu opened) :
image

Then a label is set only when the user is logged in and not displaying the mobile view, because in this case the main navigation is "splited" across multiple `<nav>` elements.

Result with Orca when the user is logged :
image

trasher: OK, those results seems perfectly correct to me; thanks :)

galette/templates/default/page.html.twig
gagnieray: > > (of course, it also can be moved elsewhere if you have a better idea :D)
> 
> You'll judge if it's better, but yes I have an other idea 😉

I moved it on the top left corner of the the screen, as the first link in the navigation (or in the banner for the mobile menu) with a pulsating transition.

On mobile : 
Capture d’écran de 2023-11-20 21-42-23

Logged in with the default menu :
Capture d’écran de 2023-11-20 21-40-49

Logged in with the compact menu :
Capture d’écran de 2023-11-20 21-41-25

trasher: Well... That's not a so bad idea; but I'm not fan, sorry :/

It make looks like there is something important happening, but in fact that only means there is an mailing in session that has not been stored... And that is not really a feature, but a kind of technical fact.

The existing mailing button is really not something important. Maybe it was at the very beginning, but now users cans store their current mailing in database and retrieve it later; disregarding the session.

For me, this button can also just be removed.

gagnieray: > For me, this button can also just be removed.

All right! I agree :+1: This non-feature is confusing :sweat_smile:

trasher (approved):

PR #371, opened 2023-11-28
Feature/fix/1748

trasher:


trasher:
As far as I've seen, this fix issues I've recently introduced; but in some scrren configuration, I continue having an hidden overflow:

image

gagnieray (approved):
:+1: but there will always be cases where the table will be larger than the screen. In such cases it is necessary to control the overflow from the table's container to allow horizontal scrolling : https://github.com/galette/galette/pull/372

PR #372, opened 2023-11-28
Fix large tables overflow when screen is to small; fixes #1748

gagnieray:


trasher:
I've retrieved the commit you made, and it's applied on develop branch; I should have merge this one first (I did not pay attention)

PR #373, opened 2023-11-30
Take care of reminders sent last period

trasher:


galette/lib/Galette/Repository/Reminders.php
gagnieray, line 1: Unfortunately these improvements alone change nothing on my side (whether membership is set as a date or as a period) :disappointed:

In both cases, the only reminder sent the next period is still the late reminder after more than 60 days. 

The impending reminders and the late reminder after more than 30 days are still never sent.

In my opinion the problem relies in the select query statement (at least regarding the impending reminders) :

https://github.com/galette/galette/blob/402e9e7971076ff80e7da3c0c87e33a7cc1e31d2/galette/lib/Galette/Repository/Reminders.php#L122-L123

If I'm not wrong, this only selects the very last late reminder sent for each member.
Impending reminders are never selected, and thus, are never treated after in the next period.

Modifying those lines into the following changes things a bit :
 
```php
'last_reminder' => 'reminder_date',
'reminder_type' => 'reminder_type'
```

Then, on the next periods, both impending reminders are properly send to concerned members.

**But**, active members late for more than 60 days will continue to receive both late reminders _as long as their account is not set to inactive_. 

Which is not the supposed expected behaviour, I think. Correct me if I'm wrong but the late reminder after more than 60 days is supposed to be the last reminder sent to a member. Late members are not supposed to continue to receive late reminders on the next periods even if their account is still active, are they ?

gagnieray: > Modifying those lines into the following changes things a bit :
> 
> ```
> 'last_reminder' => 'reminder_date',
> 'reminder_type' => 'reminder_type'
> ```

This change has another drawback... The reminders are sent each day of the impending or late periods :sweat_smile:

trasher:
OK, I think I'll give up for 1.0.0; this probably need to be entirely rewritten...

galette/lib/Galette/Repository/Reminders.php
trasher: > Unfortunately these improvements alone change nothing on my side (whether membership is set as a date or as a period) 😞

That's a bit surprising nothing changes, because it does in tests :/

> In my opinion the problem relies in the select query statement (at least regarding the impending reminders) :
> [...]

I don't think so because the type and date can be reset later in the PHP code: this information is important in regular basic cases, and reset in other ones.

> **But**, active members late for more than 60 days will continue to receive both late reminders _as long as their account is not set to inactive_.
> [...]

I think that's the issue I was talking about in my latest comment on the bug tracker :( All late members will receive a >=60 days late reminder at each new period; while they should indeed not.
But since the rest of the code is not OK, I'm not going to spend time on that. I prefer the current behaviour where nothing is sent rather than "spamming" members for now.

gagnieray: > That's a bit surprising nothing changes, because it does in tests :/

It doesn't seem that existing reminders from previous membership periods are taken in account ; only existing ones for the current membership period are added sometimes during the tests :thinking: 

> > In my opinion the problem relies in the select query statement (at least regarding the impending reminders) :
> > [...]
> 
> I don't think so because the type and date can be reset later in the PHP code: this information is important in regular basic cases, and reset in other ones.

I inspected the result set at different steps of the query's building, and definitely, after the join clauses, **when late reminders from previous periods exist in the database**, because of `new Expression('MAX(...)')`, all the result set values (date and type) are replaced with the values of the last late reminder stored in the database.

> > **But**, active members late for more than 60 days will continue to receive both late reminders _as long as their account is not set to inactive_.
> > [...]
> 
> I think that's the issue I was talking about in my latest comment on the bug tracker :( All late members will receive a >=60 days late reminder at each new period; while they should indeed not. 
> But since the rest of the code is not OK, I'm not going to spend time on that. I prefer the current behaviour where nothing is sent rather than "spamming" members for now.

Is there a way to count the number of existing late reminders of the members, and if there are more than 2, not to add those members' ids to the wave of reminders to be sent ?

trasher:
I've reworked the reminders load system; I had headache trying to change old queries.

Rather than loading late/impending members from reminders table; I choose another approach:
1- I use members filtering system to retrieve late or impending members (just as it's done UI side),
2- For each selected member, I check for an existing reminder of the same type during current period
3a- if there is no existing reminder; I just check current date is after first expected reminder one;
3b- if there is an existing reminder; I check its date against current date and found reminder one.

The logic seems (to me) more simple, therefore the code is also more simple :)
Existing unit tests are still working without changes; since I've not been able to reproduce the real issue in tests; I guess it's a good point...

Can you please check if this new proposal seems better?

Thanks!

trasher:
I don't know if original issue is resolved, but new code is more easy to understand and debug; and several tests has been added; so I'll merge this one anyway - I do not think this will be worse :)

PR #375, opened 2023-12-02
Fix reminders select query statement; closes #1614

gagnieray:


gagnieray:
> Now I need to have a look at the unit tests and understand why they are failing 😕

Well... The tests are failing rightly :disappointed:

With such a select query, this can only works when reminders are already existing in the database. Or in other words, this cannot work on a clean database.

Another interesting point : MySQL and PostgreSQL are throwing error that MariaDB doesn't : 
https://github.com/galette/galette/actions/runs/7072673252/job/19251767760?pr=375
https://github.com/galette/galette/actions/runs/7072673252/job/19251767994?pr=375
https://github.com/galette/galette/actions/runs/7072673252/job/19251767857?pr=375

trasher:
> [...]
> Another interesting point : MySQL and PostgreSQL are throwing error that MariaDB doesn't : [...]

Oh :( 

PostgreSQL always have failed as expected. For MySQL and MariaDB, with default configuration, only a Warning is emitted; therefore an extra check is done in tests:
- https://github.com/galette/galette/blob/develop/tests/GaletteTestCase.php#L151
- https://github.com/galette/galette/blob/develop/galette/lib/Galette/Core/Db.php#L1032

Looks like this has changed somehow in MariaDB; I'll probably have to add a test on that `getWarnings()` method.

trasher:
> > Now I need to have a look at the unit tests and understand why they are failing 😕
> 
> Well... The tests are failing rightly 😞

OK, well; but appart of that, would you be able to add a test case that is currently fixed by your proposal? Since #373 did not fix your issue; I guess my new test was wrong and I still have no idea how to reproduce...

trasher:
I keep this one open so I can take a look on it while trying to fix remaning cases

trasher:
Since I've completely remove the "max" reminder type part of the query, this one is no longer needed.

The only remaining change is the `a.date_echeance IS NOT NULL` clause; but since new system relies on standard impending/late filters in use for years in the members list, I guess this case is not needed neither :)

PR #383, opened 2023-12-14
Fix email settings test in preferences

gagnieray:
- don't use members administrator's emails if set in preferences, as testing doesn't accept multiple recipients
- use FUI toast message on success

### DIFF
--- a/galette/templates/default/pages/preferences.html.twig
+++ b/galette/templates/default/pages/preferences.html.twig
@@ -490,7 +490,7 @@
                         </div>
                         <br/>
                         <a
-                            href="{{ url_for('testEmail') }}#mail"
+                            href="{{ url_for('testEmail') }}"
                             id="btnmail"
                             class="ui labeled icon button"
                         >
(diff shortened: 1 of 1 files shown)


galette/templates/default/pages/preferences.html.twig
```diff
@@ -921,8 +921,7 @@
                 $('#btnmail').on('click', function(e) {
                     e.preventDefault();
                     var _this = $(this);
-                    var _value = $('#pref_email_newadh').val();
-                    var _input = '<div class="ui input"><input type="text" name="email_adress" id="email_adress" value="' + _value + '"/></div>';
+                    var _input = '<div class="ui input"><input type="text" name="email_adress" id="email_adress" value=""/></div>';
```
trasher, line 924: An alternative would be to use something like:
```
_value.split(',', 1)[0]
```

gagnieray: I don't know why, but doing that, the modal is never displayed and the test message sent directly. 

When the "members administrator's emails" field is not filled, it is not possible to choose a custom recipient in the modal and sending the test message fails.

trasher: OK, if all is OK with the field empty; it's enough for me ;) thanks!

trasher (approved):

PR #386, opened 2023-12-25
Feature/typehinting

trasher:


trasher:
Fixes https://bugs.galette.eu/issues/1766

PR #391, opened 2023-12-29
Fix/1764

trasher:
Another proposal to fix issue 1764; replacing  #388 and #389

Using code to concatenate a string, and then another piece of code to split the concatenated string seems a bit dumb (and causes issues).

My proposal is to use a new method that returns directly all parents groups names as an array (https://github.com/galette/galette/commit/fd02a4e6b14b67cb8de2f8d9f434745a141e7b67).

Also, indeed display with a "/" is impossible to understand currently. In my mind, using a FUI breadcrumb was the solution (see commit https://github.com/galette/galette/commit/4149c9043a62b7ebfd49219bae520fc3db30138e), but display is not OK at the moment (I'm not sure I did things right; it may requires some changes in FUI Galette config).

Other commits are minor changes.

Note for myself: unit tests must be added on the new method.

### DIFF
--- a/galette/templates/default/elements/group.html.twig
+++ b/galette/templates/default/elements/group.html.twig
@@ -5,12 +5,11 @@
 
 <div class="ui stackable pointing inverted menu tabbed">
     <div class="item header">
-        <i class="users cog icon" aria-hidden="true"></i>
         {{ group.getFullName() }} :
     </div>
     <a href="{{ url_for('groups', {'id': group.getId()}) }}?tab=group_information" class="item{{ tab == 'group_information' ? ' active' }}" data-tab="group_information">{{ _T("Information") }}</a>
-    <a href="{{ url_for('groups', {'id': group.getId()}) }}?tab=group_managers" class="item{{ tab == 'group_managers' ? ' active' }}" data-tab="group_managers">{{ _T("Managers") }} ({{ managers|length }})</a>
-    <a href="{{ url_for('groups', {'id': group.getId()}) }}?tab=group_members" class="item{{ tab == 'group_members' ? ' active' }}" data-tab="group_members">{{ _T("Members") }} ({{ members|length }})</a>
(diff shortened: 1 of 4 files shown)


trasher:
> Also, indeed display with a "/" is impossible to understand currently. In my mind, using a FUI breadcrumb was the solution (see commit [4149c90](https://github.com/galette/galette/commit/4149c9043a62b7ebfd49219bae520fc3db30138e)), but display is not OK at the moment (I'm not sure I did things right; it may requires some changes in FUI Galette config).

This also can be fixed with another FUI component; or changing the separator as proposed in #388

galette/templates/default/elements/group.html.twig
```diff
@@ -5,12 +5,17 @@
 
 <div class="ui stackable pointing inverted menu tabbed">
     <div class="item header">
-        <i class="users cog icon" aria-hidden="true"></i>
-        {{ group.getFullName() }} :
+        <div class="ui inverted breadcrumb">
+            {% for parent in group.parents %}
+                <a class="section">{{ parent }}</a>
```
gagnieray, line 10:
```suggestion
                <span class="section">{{ parent }}</span>
```

galette/templates/default/elements/group.html.twig
```diff
@@ -5,12 +5,17 @@
...
+                <i class="right angle icon divider"></i>
+            {% endfor %}
+            <a class="section active">{{ group.getName() }}</a>
```
gagnieray, line 13:
```suggestion
            <span class="section active">{{ group.getName() }}</span>
```

gagnieray: The `active` class here has no effect because of the `item header` container on line 7 : the whole breadcrumb content is bold.

galette/templates/default/elements/group.html.twig
gagnieray: Parents are inverted : 
Capture d’écran de 2023-12-29 20-16-54

gagnieray (changes requested):

PR #393, opened 2023-12-30
Suppress warning if open_basedir restricts access; closes #1765

trasher:


trasher:
Waiting for reporter to confirm the fix

PR #394, opened 2023-12-30
Add test on contribution next year begin date

trasher:
refs #1762


trasher:
@gagnieray could you take a look and tell me if that proposition seems OK? Thanks :)

gagnieray (approved):
> @gagnieray could you take a look and tell me if that proposition seems OK? Thanks :)

Sorry for the delayed answer.
I approve this solution :+1:

trasher:
Great, thanks for the feedback! :)

PR #425, opened 2024-02-15
Add hyperling of the asso website to the footer

TheFab21:


trasher:
I'm sorry but I do not agree with this change.

Footer text may contain HTML, we cannot know which tags will be used; so we cannot make the whole footer text in a `a` tag.
Plus, it may not be relevant to link asso website here for some cases. Since footer can be HTML, website URL can just be part of it, no need for further changes Galette side.

TheFab21:
Hi trasher, I understand, you're totaly right, my fault (i've put the website as footer, but we can put what ever we want)

PR #426, opened 2024-02-15
Add hyperlink of asso website to the logo on the sidebar menu

TheFab21:


galette/templates/default/elements/navigation/navigation_aside.html.twig
```diff
@@ -25,7 +25,9 @@
 
 {% if not login.getCompactMenu() %}
     <div class="ui basic center aligned fitted segment">
+<a href="https://{{ preferences.pref_footer|raw }}">
```
trasher, line 28: I guess you meant:
```suggestion
<a href="https://{{ preferences.pref_website|raw }}">
```

But also, we certainly need to check if this value is not empty

PR #427, opened 2024-02-16
New version checking

trasher:


trasher:
image

trasher:
Since I rely on flash messages; message is only displayed on the second loaded page... But as it's permanent and diplsayed on all pages once logged; this is not really an issue I guess.

PR #440, opened 2024-02-21
Feature/fix/1775

gagnieray:


trasher:
Not tested yet; I just took an eye quickly on the the changes - I'll test asap.

I'm not convinced for columns configuration on each dynamic field: I'm not sure this is really useful, and this is specific for dynamic fields.
If such a feature would really be implemented, I guess this would be in a more generic way including dynamic and core fields (and that's not possible for the moment - I hope I'll find enough courage to work on it for the next+ major release, this will requires a lot of changes).

As of the dynamic field information place, if we choose a 2 or 3 columns, with an information at the top; won't that cause display issues (as you pointed on the related ticket)?

gagnieray:
> Not tested yet; I just took an eye quickly on the the changes - I'll test asap.

I just forced pushed my last changes.
It's not yet finished, but it's functional :smile: 

> I'm not convinced for columns configuration on each dynamic field: I'm not sure this is really useful, and this is specific for dynamic fields. If such a feature would really be implemented, I guess this would be in a more generic way including dynamic and core fields (and that's not possible for the moment - I hope I'll find enough courage to work on it for the next+ major release, this will requires a lot of changes).

As explained on the tracker, I think it can be really usefull to improve the form display.
I added the same properties on both types of fields. 
It behaves exactly the same way for both.
 
> As of the dynamic field information place, if we choose a 2 or 3 columns, with an information at the top; won't that cause display issues (as you pointed on the related ticket)?

I moved this setting from the preferences to each dynamic field. That way, it can only be used on specific fields when really necessary (in some cases the position may preval over fields' alignement).

galette/templates/default/components/dynamic_fields.html.twig
```diff
@@ -72,6 +75,9 @@
                     {% endfor %}
                 </select>
             {% elseif get_class(field) == 'Galette\\DynamicFields\\Date' %}
+                {% if preferences.pref_field_information_position == 'above' %}
```
trasher, line 78:
```suggestion
                {% if field.hasInformationAbove() %}
```

galette/templates/default/components/dynamic_fields.html.twig
```diff
@@ -85,7 +91,9 @@
                         />
                     </div>
                 </div>
-                <span class="exemple">{{ _T("(yyyy-mm-dd format)") }}</span>
+                {% if preferences.pref_field_information_position == 'below' %}
```
trasher, line 94:
```suggestion
                {% if not field.hasInformationAbove() %}
```

galette/templates/default/components/dynamic_fields.html.twig
```diff
@@ -75,7 +75,7 @@
                     {% endfor %}
                 </select>
             {% elseif get_class(field) == 'Galette\\DynamicFields\\Date' %}
-                {% if preferences.pref_field_information_position == 'above' %}
+                {% if field.hasInformationAbove() %}
                     <div class="exemple">{{ _T("(yyyy-mm-dd format)") }}</div>
                 {% endif %}
```
gagnieray, line 80: Date format should actually be moved in the input placeholder like in contributions filters and advanced search.

And the translatable string should idealy be changed from `(yyyy-mm-dd format)` to `yyyy-mm-dd format` or even only `yyyy-mm-dd`
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder

trasher: Well.. Using placeholder is a good idea, but when a date is already present in the field, the user cannot know what format is required.
There is a date picker on all date fields; yes, but some may prefer to work from keyboard only.

I wonder if the "example" span should not be left as is (we can still add placeholder, I'll work on translations once current PR has been merged).

trasher: I've created draft PR #451 so I won't forget that :)

galette/install/scripts/sql/upgrade-to-1.10-mysql.sql
```diff
@@ -8,4 +8,9 @@ ALTER TABLE galette_transactions ADD FOREIGN KEY (type_paiement_trans)
     REFERENCES galette_paymenttypes (type_id)
     ON DELETE RESTRICT ON UPDATE RESTRICT;
 -- Add field_min_size to galette_field_types
-ALTER TABLE galette_field_types ADD field_min_size integer(10) NULL DEFAULT NULL
\ No newline at end of file
+ALTER TABLE galette_field_types ADD field_min_size integer(10) NULL DEFAULT NULL
```
trasher, line 11:
```suggestion
ALTER TABLE galette_field_types ADD field_min_size integer(10) NULL DEFAULT NULL;
```

galette/templates/default/components/dynamic_fields.html.twig
trasher: > Don't you think the date already present is an _actual example_ of the required date format ? 🤔

It's a good point indeed... Maybe am I just too much used to this display? Let's go for placeholder only display, we can adjust quite simply in the future if required ;)

galette/lib/Galette/Entity/FieldsConfig.php
```diff
@@ -344,14 +345,15 @@ function ($field) use (&$_all_fields) {
...
-                            'list_position' => $f['list_position'] ?? null
+                            'field_id'       => $k,
+                            'table_name'     => $this->table,
+                            'required'       => $f['required'],
+                            'visible'        => $f['visible'],
+                            'position'       => $f['position'],
+                            'category'       => $f['category'],
+                            'list_visible'   => $f['list_visible'] ?? false,
+                            'list_position'  => $f['list_position'] ?? null,
+                            'width_in_forms' => $f['width_in_forms'] ?? null
```
trasher, line 356:
```suggestion
                            'width_in_forms' => $f['width_in_forms'] ?? 1
```

trasher (approved):

PR #455, opened 2024-03-03
Asso documents; fixes #1216

trasher:


trasher:
I have to add tests, but it''s mostly ready.

@gagnieray could you please take an eye on the public list part; from a design PoV? Management part is quite simple and standard, but I've tried to do something different for the public page.
Thanks!

gagnieray:
> @gagnieray could you please take an eye on the public list part; from a design PoV? Management part is quite simple and standard, but I've tried to do something different for the public page. Thanks!

I'll have a look as soon as possible ! :smile:

PR #457, opened 2024-03-05
Fix preferences tabs display on small screens

gagnieray:
fixes #1799


trasher (approved):

galette/templates/default/pages/preferences.html.twig
```diff
@@ -22,7 +22,7 @@
 
 {% block content %}
         <form action="{{ url_for('store-preferences') }}" method="post" enctype="multipart/form-data" class="ui form">
-        <div class="ui stackable pointing inverted menu tabbed">
+        <div class="ui stackable pointing inverted wrapping menu tabbed">
```
trasher, line 25: Waht about?
```suggestion
        <div class="ui stackable pointing inverted wrapping centered fluid menu tabbed">
```

trasher: I've applied since I prefer this display (and all seems OK) :)

gagnieray: The "drawback" of this variation is that it applies also on large screens, making tabs wider.

Default:
large-default-preferences

Centered fluid:
large-centered-fluid-preferences

For UI consistancy, shouldn't we use it on every screen ?
But with few tabs only, I personnaly don't find that great and I'd prefer the tabs to remain "compact".

Default:
large-default-few-tabs

Centered fluid:
large-centered-fluid-few-tabs

It's up to you ! :wink:

And I suppose that if it is used on preferences only, it won't break the UI consistancy that much :grin:

trasher: > The "drawback" of this variation is that it applies also on large screens, making tabs wider.
> [...]
> It's up to you ! 😉
> 
> And I suppose that if it is used on preferences only, it won't break the UI consistancy that much 😁

Indee I did no think about other tabs usage; but preferences is the only one that needs wrapping; and in that cas the fluid fits correctly. I'll keep it; it's easy to change anyway :smile:

PR #514, opened 2024-06-04
Add e2e tests

trasher:
Trying to add tests with selenium, phpunit, browserstack on github actions

- First..; well, it does not work right now :rofl: 
- Second, current existing test suite takes very much longer to run (approx 1"30 for unit tests, becames +6"00). Maybe a timeout issue since it does not work right now?

Maybe mixing "standard" phpunit tests wih e2e ones is not a good ide, I do not know yet.


trasher:
That would be really great to have e2e tests; but it would take too much time; and I have absolutely no experience. Until I got some help to setup that, I'll not work on it.

PR #515, opened 2024-06-09
Hotfix/1.1.1

trasher:


trasher:
Replaced by https://github.com/galette/galette/pull/516

PR #537, opened 2024-09-19
Update PdfMembersCards.php

TheFab21:
Update PdfMembersCards.php
correct pb with landscape logos
adaptive fontsize  for the banner

### DIFF
--- a/galette/lib/Galette/IO/PdfMembersCards.php
+++ b/galette/lib/Galette/IO/PdfMembersCards.php
@@ -64,6 +64,7 @@ class PdfMembersCards extends Pdf
     private float $wlogo;
     private float $hlogo;
     private string $logofile;
+    private int $nb_car;
 
     /**
      * Main constructor, set creator and author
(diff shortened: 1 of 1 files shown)


galette/lib/Galette/IO/PdfMembersCards.php
```diff
@@ -308,13 +309,17 @@ public function drawCards(array $members): void
...
+				$nb_car = 12 ;
+				}
             $this->SetFillColor($fcol['R'], $fcol['G'], $fcol['B']);
             $this->SetTextColor(
                 $this->tcol['R'],
                 $this->tcol['G'],
                 $this->tcol['B']
             );
-            $this->SetFont(self::FONT, 'B', 8);
+            $this->SetFont(self::FONT, 'B', $nb_car);
```
trasher, line 322: Well, I'm not sure at all since the value is either arbitrary; or calculated a way I do not understand.

trasher:
See #535; which include fix provided on discord channel. To enable adaptative cards, the `GALETTE_ADAPTATIVE_CARDS` must be set to @true@ in `galette/config/behavior.inc.php`.

PR #538, opened 2024-09-20
fix(ScheduledPayment) fix error when posting scheduled payment

3bandiste:
When saving a scheduled payment on a contribution, $data[Contribution::PK] is a string and we get an error as it should be casted as an int.
I also fixed the expected name for the `paid` parameter which is not `is_paid` but `paid`

### DIFF
--- a/galette/lib/Galette/Entity/ScheduledPayment.php
+++ b/galette/lib/Galette/Entity/ScheduledPayment.php
@@ -197,7 +197,7 @@ class ScheduledPayment
             ->setPaymentType((int)$data['id_paymenttype'])
             ->setCreationDate($data['creation_date'] ?? date('Y-m-d'))
             ->setScheduledDate($data['scheduled_date'])
-            ->setAmount((float)$data['amount'] ?? $this->contribution->amount)
+            ->setAmount(isset($data['amount']) ? (float)$data['amount'] : $this->contribution->amount)
             ->setPaid($data['paid'] ? boolval($data['paid']) : false)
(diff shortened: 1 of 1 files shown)


galette/lib/Galette/Entity/ScheduledPayment.php
```diff
@@ -193,12 +193,12 @@ public function check(array $data): bool
...
         $this
-            ->setContribution($data[Contribution::PK])
-            ->setPaymentType($data['id_paymenttype'])
+            ->setContribution((int)$data[Contribution::PK])
+            ->setPaymentType((int)$data['id_paymenttype'])
             ->setCreationDate($data['creation_date'] ?? date('Y-m-d'))
             ->setScheduledDate($data['scheduled_date'])
-            ->setAmount($data['amount'] ?? $this->contribution->amount)
-            ->setPaid($data['is_paid'] ?? false)
+            ->setAmount((float)$data['amount'] ?? $this->contribution->amount)
```
trasher, line 200: I guess typehinting a non existing array entry will fail:
```suggestion
            ->setAmount(isset($data['amount']) ? (float)$data['amount'] : $this->contribution->amount)
```

trasher:
Thank you for the fix! All seems good, I just see a potential issue (see my proposal).

3bandiste:
> Thank you for the fix! All seems good, I just see a potential issue (see my proposal).

Thanks you're right

trasher (approved):

PR #544, opened 2024-09-30
Use identity columns for PostgreSQL instead of sequences

trasher:


galette/lib/Galette/Core/Db.php
```diff
@@ -1015,7 +1035,7 @@ public function getLastGeneratedValue(object $entity): int
         /** @phpstan-ignore-next-line */
         return (int)$this->driver->getLastGeneratedValue(
             $this->isPostgres() ?
-                PREFIX_DB . $entity::TABLE . '_id_seq'
+                PREFIX_DB . $entity::TABLE . '_' . $entity::PK . '_seq'
```
trasher, line 1038:
```suggestion
                $this->getSequenceName($entity::TABLE, $entity::PK, true)
```

trasher:
Replaced with https://github.com/galette/galette/pull/548

PR #551, opened 2024-10-10
Fixe Size problem for $an_cot in adaptativecards

TheFab21:


TheFab21:
CS corrected with last update

galette/lib/Galette/IO/PdfMembersCardsAdaptative.php
```diff
@@ -195,16 +195,24 @@ public function drawCards(array $members): void
...
 
             //Write shadow of the year, center of the card
             $an_cot = $this->an_cot;
             if ($an_cot === 'DEADLINE') {
                 //get current member deadline
                 $an_cot = $member->due_date;
             }
-
-            $xan_cot = $x0 + $this->wi / 2 - $this->GetStringWidth(
+            if (is_null($an_cot)) {
```
trasher, line 205: This condition does not work. I do not know exactly which problem you tried to fix here; but there is something wrong.

trasher (changes requested):

galette/lib/Galette/IO/PdfMembersCardsAdaptative.php
trasher: I've been able to reproduce and fix.

trasher (approved):

PR #560, opened 2024-11-06
A try with psalm security tool

trasher:


trasher:
Tool always timeout, I do not understand why... I close, I've already spent too much time on that one.

PR #580, opened 2025-01-23
Single line links in footer

gagnieray:
I started this PR because the existing "Website" link in the footer is a bit confusing for users that can think it is a link to the association's website.

It ended this way :
* the website "link" was changed in a simple "Galette" link
* the version is only displayed when a user is logged-in, or also on public pages when in demo mode (I think it's better to hide the version on public pages in production mode to harden the discovery of obsolete instances with search engines if indexation prevention is not enabled in the preferences)
* all links are displayed on a single line

### DIFF
--- a/galette/templates/default/elements/footer.html.twig
+++ b/galette/templates/default/elements/footer.html.twig
@@ -19,8 +19,8 @@
  */
 #}
 {% set new_release = callstatic('\\Galette\\Core\\Galette', 'getNewRelease') %}
-    <footer class="page-footer{% if login.getCompactMenu() %} extended"{% endif %}">
-        <div class="ui center aligned basic padded segment footer-wrapper">
+    <footer class="page-footer">
+        <div class="ui center aligned basic padded segment footer-wrapper{% if login.getCompactMenu() %} extended"{% endif %}">
             <div class="row">
                 <div class="ui horizontal bulleted link list">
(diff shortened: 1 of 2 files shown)


galette/templates/default/elements/footer.html.twig
```diff
@@ -48,23 +32,23 @@
 {% else %}
                         <i class="icon cookie bite" aria-hidden="true"></i>
 {% endif %}
-                        Galette {{ constant('GALETTE_DISPLAY_VERSION') }}
+                        Galette{% if login.isLogged() or callstatic('\\Galette\\Core\\Galette', 'isDemo') %} {{ constant('GALETTE_DISPLAY_VERSION') }}{% endif %}
```
trasher, line 35: I guess we should keep only displayed version when user is logged in; I try to keep demo up-to-date (espacially when there are security issues, of course), but I may fail ^^
```suggestion
                        Galette{% if login.isLogged() %} {{ constant('GALETTE_DISPLAY_VERSION') }}{% endif %}
```

galette/templates/default/elements/header.html.twig
```diff
@@ -19,7 +19,7 @@
  */
 #}
 {# This file contains common html headers to include for Galette Twig rendering. #}
-        <title>{% if preferences.pref_slogan != "" %}{{ __(preferences.pref_slogan) }} - {% endif %}{% if page_title != "" %}{{ page_title }} - {% endif %}Galette {{ constant('GALETTE_VERSION') }}</title>
+        <title>{% if preferences.pref_slogan != "" %}{{ __(preferences.pref_slogan) }} - {% endif %}{% if page_title != "" %}{{ page_title }} - {% endif %}Galette{% if callstatic('\\Galette\\Core\\Galette', 'isDemo') %} {{ constant('GALETTE_VERSION') }}{% endif %}</title>
```
trasher, line 22: Not sure haveing version in page title is really relevant
```suggestion
        <title>{% if preferences.pref_slogan != "" %}{{ __(preferences.pref_slogan) }} - {% endif %}{% if page_title != "" %}{{ page_title }} - {% endif %}Galette</title>
```

PR #589, opened 2025-02-02
Dynamic dates conversion from admintools

trasher:
refs #1881


galette/lib/Galette/DynamicFields/Date.php
```diff
@@ -55,4 +57,94 @@ public function getType(): int
...
+                    array(
+                        'field_val'  => $v,
+                        DynamicFieldsHandle::PK => $k
+                    )
+                );
+            }
+            throw new \RuntimeException(
+                print_r($debug_dates, true)
+            );
+            //$zdb->connection->commit();
```
trasher, line 138: To be uncommented when ready

PR #595, opened 2025-02-09
Add links to the documentation when relevant; fixes #1905

gagnieray:


galette/templates/default/page.html.twig
```diff
@@ -69,6 +62,25 @@
                             </h1>
 {% if login.isLogged() %}
                         </div>
+    {% if login.isAdmin() or login.isStaff() or login.isGroupManager() %}
+        {% set documentation_path = documentation ?? '' %}
+        {% set documentation_language = ((galette_lang == 'ota' or galette_lang == 'si') ? 'en' : galette_lang == 'nb' ? 'no' : galette_lang) %}
+        {% if documentation_path is not empty %}
+                        <div class="ui compact segment">
+                            <a class="circular ui basic icon button" href="https://doc.galette.eu/{{ documentation_language }}/master/{{ documentation }}" target="_blank" data-position="left center" title="{{ _T("Read the manual") }}">
```
trasher, line 70: I've not tested the feature yet, but I have 2 main remarks:
- pointing to `master` may lead to missing page when using a devleopment version (such as a nightly build)
- anchors are localized :/

For the first issue, that may not be a problem, but for the second, it is. The only thing we can link is a page, not a part of it

galette/webroot/installer.php
```diff
@@ -202,12 +202,19 @@
     $install->atEndStep();
 }
 
+$documentation_language = $i18n->getAbbrev();
```
trasher, line 205: A similar condition to fix locales names has also been added. I guess a method managed from the I18n object would be preferable (it should ba available on both cases).

galette/templates/default/page.html.twig
gagnieray: I think I can handle that.

galette/webroot/installer.php
gagnieray: All right!

trasher:
Seems good, thank you :+1:  :)

Two minor points:
- would it be possible to keep right alignment?
- maybe the doc icon could be a little smaller

gagnieray:
>     * would it be possible to keep right alignment?

Right alignment is preserved already.
Don't forget `npm run build`.

>     * maybe the doc icon could be a little smaller
Of course. I have applied the mini button size variation and forced push the branch one last time.

trasher:
> Don't forget `npm run build`.

Good catch, I missed that once again; sorry for the noise...

> 
> > ```
> > * maybe the doc icon could be a little smaller
> > ```
> 
> Of course. I have applied the mini button size variation and forced push the branch one last time.

:+1:

perfect, thank you :)

trasher (approved):
Sounds good, thank you :) I'll try tomorrow

PR #596, opened 2025-02-11
Minor fixes on installer and maintenance pages

gagnieray:


galette/install/steps/admin.php
```diff
@@ -40,7 +40,7 @@
 
     <div class="ui equal width grid">
         <div class="right aligned column">
-            <button type="submit" class="ui right labeled primary icon button"><i class="angle double right icon" aria-hidden="true"></i> <?php echo _T("Next step"); ?></button>
+            <button type="submit" class="ui right labeled primary icon button"><i class="angle double <?php if ($i18n->isRtl()) { ?>left<?php } else { ?>right<?php } ?> icon" aria-hidden="true"></i> <?php echo _T("Next step"); ?></button>
```
trasher, line 43: Maybe this would be easier to read?
```suggestion
            <button type="submit" class="ui right labeled primary icon button"><i class="angle double <?php echo $i18n->isRtl() ? 'left' : 'right';  ?> icon" aria-hidden="true"></i> <?php echo _T("Next step"); ?></button>
```

Looks good anyway ;)

gagnieray: > Maybe this would be easier to read?

Indeed !

trasher (approved):

PR #597, opened 2025-02-11
Improve groups selection tree UX

gagnieray:
- browse groups by clicking on their name instead of clicking on their arrow
- add an icon to select and edit a group (replaced by an open folder when group is selected)
- fix alignements in case of very long names (by removing the use of labels on names)

Before :
tree-before

After :
tree-after


trasher:
I'm OK for the change, but the edit button position does not convince me; since it depends on group name length. 

What about placing the edit icon before the group name? That way, all edit icons would be at the same place for the same level.

trasher (approved):

PR #599, opened 2025-02-11
Feature/customizable main colors

gagnieray:


gagnieray:
Documentation is ready : https://github.com/galette/galettedoc/pull/33

trasher:
Tested, everything seems OK.

I've just changed the way the CSS reset is handled from JS to PHP. Also, previous code was not aware of "Apply custom colors" state change.

gagnieray:
> I've just changed the way the CSS reset is handled from JS to PHP. Also, previous code was not aware of "Apply custom colors" state change.

Great ! This is way better :+1: :rocket:

PR #603, opened 2025-02-13
Staff public list and gallery

trasher:


trasher:
@gagnieray maybe could you help on displays for this one.

I have 2 issues I do not find how to solve:
1- on staff list, when groups managers are displayed, I've used a `<th colspan="">` to identify table sub-headers, but the display is really not good
image

2- on staff gallery, I did not find any way to separate staff and groups managers; therefore if one is part of staff and manages a group (or manage several groups); it looks like a duplicate.

gagnieray:
> I have 2 issues I do not find how to solve: 
> 1- on staff list, when groups managers are displayed, I've used a `<th colspan="">` to identify table sub-headers, but the display is really not good !

With FUI, `<th>` are not supposed to be used outside `<thead>` :confused: 
I encountered the same issue on the plugins page and I added specific paddings in CSS (and colors also).

Another solution is to replace `<th>` with `<td>` and apply the "active" class on `<tr>` : https://fomantic-ui.com/collections/table.html#active

Which solution do you prefer ?

> 2- on staff gallery, I did not find any way to separate staff and groups managers; therefore if one is part of staff and manages a group (or manage several groups); it looks like a duplicate.

I'll have a look and see what can be done in that case.

galette/templates/default/pages/public/staff_gallery.html.twig
```diff
@@ -20,20 +20,40 @@
...
 {% block infoline %}
     {# Empty this block #}
 {% endblock %}
 
-{% block body %}
+{% block groups_menu %}
     {% if nb_members > 0 %}
-        {% for member in members.staff %}
-            {{ _self.drawEntry(member, true, time) }}
+        <div class="ui stackable pointing inverted wrapping centered fluid menu tabbed">
```
gagnieray, line 31: Using an horizontal stackable menu might not be the best choice depending on the total number of groups to display :thinking: 

I have investigated in using the accordion tree menu. It might be possible to use an accordion tree as a tab menu. It would have been ideal I think. But the required data is not available to the template.

Do you think it could be possible to provide the required data to the template ?

trasher: > Do you think it could be possible to provide the required data to the template ?

What is currently missing?

trasher: Using an accordion, only one can be displayed at once, right? (like the stacked menu proposed).

It does not really make sense when there are several groups with only one manager :/

I would have preferred a solution that display all photos at once; maybe keep original display and removing duplicates would be the solution (I've added groups managed in display for each staff member)

gagnieray: > Using an accordion, only one can be displayed at once, right? (like the stacked menu proposed).

Yes.

But regarding the accordion tree, on reflexion, it is over-complicated, considering only groups with managers who accepted to be in the list should be displayed in the tree, and if none, the hierarchy should still be preserved.

A dropdown (eventually a search one when there are a lot of groups) could be a good alternative.

> It does not really make sense when there are several groups with only one manager :/

Not wrong :D
That's the drawback of separating staff from group managers this way.

> I would have preferred a solution that display all photos at once; maybe keep original display and removing duplicates would be the solution (I've added groups managed in display for each staff member)

Removing duplicates instead of separating staff and group managers would indeed be the only solution to achieve that result. But I see no way this could be done with twig in the actual state of code. De-duplication should be done by the controller.

trasher: > Removing duplicates instead of separating staff and group managers would indeed be the only solution to achieve that result. But I see no way this could be done with twig in the actual state of code. De-duplication should be done by the controller.

Yes, of course! I'll take care of that ;-)

trasher:
I guess this one is OK now. I'll try to make local tests quickly, so it can be merged.

gagnieray:
> I guess this one is OK now. I'll try to make local tests quickly, so it can be merged.

PHP errors are thrown when the rights "inherit" or "hidden" are used :

```
[Tue Feb 18 11:06:33.956209 2025] [php:notice] [pid 244688:tid 244688] [client ::1:55534] Slim Application Error
Type: Twig\\Error\\RuntimeError
Code: 0
Message: An exception has been thrown during the rendering of a template ("Unknown public pages right: 3").
File: /var/www/galette/galette/templates/default/elements/navigation/public_pages.html.twig
Line: 25

[...]

Previous Error:
Type: RuntimeException
Code: 0
Message: Unknown public pages right: 3
File: /var/www/galette/galette/lib/Galette/Core/Preferences.php
Line: 1056
Trace:
#0 /var/www/galette/galette/lib/Galette/Core/Galette.php(585): Galette\\Core\\Preferences->showPublicPage()
#1 [internal function]: Galette\\Core\\Galette::getPublicMenus()

[...]
```

trasher:
> PHP errors are thrown when the rights "inherit" or "hidden" are used :

Indeed; this should be fixed in last commit, I also added missing tests cases :)

gagnieray:
> Indeed; this should be fixed in last commit, I also added missing tests cases :)

Yes. PHP errors are gone.

But there are still issues with rights applied.
"Inherit" and "Hidden" have the same value (3) in preferences.
Thus "Inherit" cannot be used :confused:

gagnieray:
And when "Hidden" is used, pages are inaccessible to everyone. But the link to the documents public list is still displayed in the menu for superadmin.

trasher:
> But there are still issues with rights applied. "Inherit" and "Hidden" have the same value (3) in preferences. Thus "Inherit" cannot be used 😕

OK, found it: the wrong constant was used in template file. Everything  should be OK now.

> And when "Hidden" is used, pages are inaccessible to everyone. But the link to the documents public list is still displayed in the menu for superadmin.

Well, that's how it's been implemented when documents has been added. Main reason (as far as I remember) is the public documents page display does not exists elsewhere. I do not know if we should keep this particularity, but that's not stricly related to current PR :)

trasher:
Documents public menu now uses the same logic as all others; this makes things more simple :)

gagnieray:
> Documents public menu now uses the same logic as all others; this makes things more simple :)

Tested. It's working as expected :ok_hand:

PR #605, opened 2025-02-14
Slightly re-ordering items in menu

gagnieray:


galette/lib/Galette/Core/Galette.php
```diff
@@ -207,13 +207,22 @@ public static function getMenus(bool $public = false): array
...
+                        'route' => [
+                            'name' => 'addMember'
+                        ]
+                    ],
+                    [
+                        'label' => _T("Empty adhesion form"),
+                        'title' => _T("Download empty adhesion form"),
+                        'route' => [
+                            'name' => 'emptyAdhesionForm'
+                        ]
```
gagnieray, line 223: As this is not an authenticated route, what do you think about adding a link on unauthenticated pages ? (with a setting to let admin choose to display this link or not :sunglasses:)

gagnieray: > As this is not an authenticated route, what do you think about adding a link on unauthenticated pages ? (with a setting to let admin choose to display this link or not 😎)

See #606

trasher: Yes, that can make sense, thank you :)

trasher (approved):

PR #606, opened 2025-02-14
Add link to download the PDF form on public pages

gagnieray:


galette/templates/default/pages/member_form.html.twig
```diff
@@ -120,6 +120,24 @@
         </div>
         <div class="sixteen wide tablet ten wide computer eleven wide widescreen column">
     {% endif %}
+
+    {% if (member.id is empty or self_adh and members.list is not defined) and preferences.pref_bool_empty_form_link %}
```
trasher, line 124: Not sure this one is useful; link from public pages should be enough.

galette/templates/default/pages/preferences.html.twig
```diff
@@ -319,7 +319,7 @@
                 <i class="eye outline icon"></i>
                 {{ __('Public pages') }}
             </div>
-            <div class="ui stackable two column grid">
+            <div class="ui stackable three column grid">
```
trasher, line 322: I've not tested, but public relarted pages in preferences will be reworked #603; maybe would it be better to work directly with reworked page and make PR against `feature/staff-gallery` branch?

galette/templates/default/pages/member_form.html.twig
gagnieray: When self-subscription is disabled, I chose to replace the subscription link with the link to download the pdf :
pdf-in-menu

But when self-subscription is enabled, in my opinion, it is logically not necessary to provide the pdf link. Unless admins want to give the user the choice of subscribing by mail or in person (for those who are reluctant to online forms). That's why in that case I chose not to add a supplementary link in the menu, but rather a message above the registration form :
pdf-in-message

galette/templates/default/pages/preferences.html.twig
gagnieray: > I've not tested, but public relarted pages in preferences will be reworked #603; maybe would it be better to work directly with reworked page and make PR against `feature/staff-gallery` branch?

All right ! :+1:

galette/templates/default/pages/member_form.html.twig
trasher: OK, I did not understood correctly the goal; thanks for the explanation. Seems OK then ;)

gagnieray:
New PR made against `feature/staff-gallery` branch : #607

PR #608, opened 2025-02-16
Feature/public pages dropdowns

gagnieray:
Group public pages in dropdowns when logged out to give space in the navigation top bar.
public-pages-dropdowns

If all public pages are not accessible, simple items are used as usual.
public-pages-staff-only

PR #609, opened 2025-02-19
Public pages dropdowns

trasher:
Group public pages in dropdowns when logged out to give space in the navigation top bar.
public-pages-dropdowns

If all public pages are not accessible, simple items are used as usual.
public-pages-staff-only

Replaces https://github.com/galette/galette/pull/608

### DIFF
--- a/galette/lib/Galette/Core/Galette.php
+++ b/galette/lib/Galette/Core/Galette.php
@@ -537,8 +537,11 @@ class Galette
         $menus = [];
         $items = [];
 
+        $galleries = [];
+        $lists = [];
+
         if ($preferences->showPublicPage($login, 'pref_publicpages_visibility_memberslist')) {
-            $items[] = [
+            $lists[] = [
                 'label' => _T("Members"),
                 'route' => [
                     'name' => 'publicMembersList'
@@ -548,33 +551,55 @@ class Galette
         }
 
         if ($preferences->showPublicPage($login, 'pref_publicpages_visibility_membersgallery')) {
-            $items[] = [
+            $galleries[] = [
                 'label' => _T('Gallery'),
                 'route' => [
                     'name' => 'publicMembersGallery'
                 ],
-                'icon' => 'user friends'
+                'icon' => 'images'
             ];
         }
 
         if ($preferences->showPublicPage($login, 'pref_publicpages_visibility_stafflist')) {
-            $items[] = [
+            $lists[] = [
                 'label' => _T("Staff"),
                 'route' => [
                     'name' => 'publicStaffList'
                 ],
-                'icon' => 'address card'
+                'icon' => 'address book outline'
             ];
         }
 
         if ($preferences->showPublicPage($login, 'pref_publicpages_visibility_staffgallery')) {
-            $items[] = [
+            $galleries[] = [
                 'label' => _T('Staff gallery'),
                 'route' => [
                     'name' => 'publicStaffGallery'
                 ],
-                'icon' => 'user cog'
+                'icon' => 'images outline'
             ];
+        }
+
+        if (count($lists) > 1) {
+            //handle multiple lists
+            $items[] = [
+                'label' => _T("Directories"),
+                'icon' => 'address book',
+                'children' => $lists
+            ];
+        } else {
+            $items = array_merge($items, $lists);
+        }
+
+        if (count($galleries) > 1) {
+            //handle multiple galleries
+            $items[] = [
+                'label' => _T("Galleries"),
+                'icon' => 'images',
+                'children' => $galleries
+            ];
+        } else {
+            $items = array_merge($items, $galleries);
         }
 
         if ($preferences->showPublicPage($login, 'pref_publicpages_visibility_documents')) {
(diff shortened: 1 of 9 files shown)


trasher:
There were issues when some menus were disabled; and it was quite difficult to understand why... So I've proposed an alternative (I hope more simple to understand) change.
I've quickly adapted template files to get rid of errors; final result seems to be the same (no need for `alternative` for label and icon.

Sonarcloud is complaining about the use of `@tabindex` on div elements, see https://sonarcloud.io/organizations/galette/rules?open=Web%3AS6845&rule_key=Web%3AS6845
I've removed them, but if that was on purpose we can add them back and ignore Sonarcloud complaint :)

gagnieray (commented):
> There were issues when some menus were disabled; and it was quite difficult to understand why... So I've proposed an alternative (I hope more simple to understand) change. I've quickly adapted template files to get rid of errors; final result seems to be the same (no need for `alternative` for label and icon.

Great ! That is indeed much more simple :+1: 

Your changes only required a fix for the "compact" menu. Otherwise, everything works as expected :smile: 

> Sonarcloud is complaining about the use of `@tabindex` on div elements, see https://sonarcloud.io/organizations/galette/rules?open=Web%3AS6845&rule_key=Web%3AS6845 I've removed them, but if that was on purpose we can add them back and ignore Sonarcloud complaint :)

Yes I saw that complaint. But the use of tabindex was indeed on purpose. Without, keyboard navigation is broken.

I assume in that case it is not very important as the "sidebar" menu is only displayed on small screen devices like mobile phones. It will only be a problem on desktop if the web browser window's width is narrow. 

FYI the same issue will be raised again if/when the code of the "aside" accordion menu is modified. Then in that case, tabindex should remain in place to preserve keyboard navigation, as "aside" is the default menu on desktop.

In fact, I did that initialy to reproduce what FUI does on dropdowns initialized with javascript (which is a behaviour FUI doesn't use on accordions, and that makes me wonder if accordions are supposed not to be used as navigation elements :thinking:).

Moreover, reading sonarcloud details, I learnt that using tabindex on non interactive elements can be an accessibility issue...

So, I don't know what to decide regarding the tabindex you removed.

Given these explanations, what is your opinion ?

gagnieray:
Another detail : icons.
I suggest other icons with a look closer to the members ones.

- list : `address book outline` 
  list

- gallery : `users cog` 
  gallery

trasher:
> [...]
> I assume in that case it is not very important as the "sidebar" menu is only displayed on small screen devices like mobile phones. It will only be a problem on desktop if the web browser window's width is narrow.

Well, outside testign during development, this should not happen.

> [...]
> Moreover, reading sonarcloud details, I learnt that using tabindex on non interactive elements can be an accessibility issue...

Event if it's not sure, I prefer to avoid such cases because we cannot test :/

> [...]
> Given these explanations, what is your opinion ?

As far as I understand, with current code, tabindex are not really needed for "real life" cases; but *could* cause issues. I vote to remove them. If there are issues, we can always add them back later.

>     * list : `address book outline`

OK for this one.

>     * gallery : `users cog`

Hum... `cog` should be used for users that can configure something (staff/admin). I propose (see last commit) `images outline`:

image

This way, we have "outline" icons on dropdowns, and "standard" ones for standalone pages:
image

image

gagnieray:
> > [...]
> > I assume in that case it is not very important as the "sidebar" menu is only displayed on small screen devices like mobile phones. It will only be a problem on desktop if the web browser window's width is narrow.
> 
> Well, outside testing during development, this should not happen.

I agree.

 
> > [...]
> > Moreover, reading sonarcloud details, I learnt that using tabindex on non interactive elements can be an accessibility issue...
> 
> Event if it's not sure, I prefer to avoid such cases because we cannot test :/

Fine !

 
> > [...]
> > Given these explanations, what is your opinion ?
> 
> As far as I understand, with current code, tabindex are not really needed for "real life" cases; but _could_ cause issues. I vote to remove them. If there are issues, we can always add them back later.

As I said, the only "real life" case where it is needed is on the "aside" accordion menu (the default menu when logged in). Removing them *all* will break keyboard navigation in this case for sure.

Thus, nothing more needs to be done for the moment I think.

> [...] 
> I propose (see last commit) `images outline`:
> 
> [...]
> This way, we have "outline" icons on dropdowns, and "standard" ones for standalone pages:

My suggestion concerned in fact icons for standalone pages :laughing: 

I like `images` icon for galleries though ! :+1: 

Have a look at my counter proposal (last commit) : use outline versions for staff standalone pages, and otherwise use standard versions for members standalone pages *and* dropdowns.

- "aside" menu (logged in) :
  aside

- "sidebar" menu (mobile + logged in) :
  mobile-loggedin

- "topbar" menu (logged out) in various configurations :
  top-dropdowns
  top-directories-dropdown
  top-galleries-dropdown
  top-members-only
  top-staff-only

- "sidebar" menu (mobile + logged out) in various configurations :
  mobile-dropdowns
  mobile-directories
  mobile-galleries
  mobile-members-only
  mobile-staff-only

PR #612, opened 2025-03-01
Staff gallery fixes

trasher:
Fixes for:
- https://bugs.galette.eu/issues/1911
- https://bugs.galette.eu/issues/1910

### DIFF
--- a/galette/includes/fields_defs/members_fields.php
+++ b/galette/includes/fields_defs/members_fields.php
@@ -250,7 +250,7 @@ $members_fields = array(
         'category' => FieldsCategories::ADH_CATEGORY_GALETTE
     ),
     'bool_display_info' => array(
-        'label'    => _T("Be visible in the members list:"),
+        'label'    => _T("Be visible on public pages:"),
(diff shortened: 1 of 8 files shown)


gagnieray (commented):
Regarding #1911, I just find it odd that the staff icon is also used for members who are only group managers. Also, depending on the custom statuses of members of each association, it might be an issue to publicly display the name of the status of members who are not staff members.

Could the priority of the statuses be used in this case to not display the status of the member?

trasher:
> Regarding #1911, I just find it odd that the staff icon is also used for members who are only group managers.

Hum... I must admin I'm a bit lost in icons :D I'll look at that.

> Also, depending on the custom statuses of members of each association, it might be an issue to publicly display the name of the status of members who are not staff members.
> 
> Could the priority of the statuses be used in this case to not display the status of the member?

I've fixed that; the condition to display status now depends on `Adherent::isStaff()`; this is more reliable.

gagnieray:
It's all good from my point of view :+1: :rocket:

PR #613, opened 2025-03-03
Fix tip icon size in input template file

gagnieray:


trasher (approved):
Seems good but I did not found where it is used :)

gagnieray:
> Seems good but I did not found where it is used :)

It is not used in core at the moment. I noticed it because I'm working on a small custom plugin I may soon propose to the community :smile:

PR #614, opened 2025-03-03
Feature/additions for plugins

gagnieray:
At the moment, plugins cannot add javascript code on every page. 
And the default configuration of summernote editor may not be suited to a plugin's needs.

These commits intend to change that.

### DIFF
--- a/galette/templates/default/elements/scripts.html.twig
+++ b/galette/templates/default/elements/scripts.html.twig
@@ -229,7 +229,9 @@
         <script type="text/javascript" src="{{ base_path() }}/assets/js/summernote.min.js"></script>
         <script type="text/javascript" src="{{ base_path() }}/assets/js/lang/summernote-{{ i18n.getID()|replace({'_': '-'}) }}.min.js"></script>
         <script type="text/javascript">
-            function activateHtmlEditor(elt, basic) {
+            function activateHtmlEditor(elt, basic, styletags, insert) {
+                var _styletags = ['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
+                var _insert = ['link', 'picture'];
                 if (basic === true) {
                     var _toolbar = [
                         ['font', ['bold', 'italic', 'underline', 'strikethrough', 'clear']],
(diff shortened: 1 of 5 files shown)


trasher:
> At the moment, plugins cannot add javascript code on every page.

Plugins can add scripts on every page; but in headers. Adding them after Galette scripts at the end of the page may be a good idea.

> And the default configuration of summernote editor may not be suited to a plugin's needs.

Hum... I guess you have that need so I'm OK.

galette/templates/default/page.html.twig
```diff
@@ -95,5 +95,12 @@
         </a>
         {% include "elements/scripts.html.twig" %}
         {% block javascripts %}{% endblock %}
+        {# If some additional scripts should be added from plugins, we load the relevant template file
+           We have to use a template file, so Twig will do its work (like replacing variables). #}
+        {% if plugin_scripts|length != 0 %}
+            {% for mid, script in plugin_scripts %}
+                {% include script with {'module_id': mid} %}
+            {% endfor %}
+        {% endif %}
```
trasher, line 104: Should not that be part of the "elements/scripts.html.twig"? Would prevent to add you code twice :)

galette/templates/default/elements/scripts.html.twig
```diff
@@ -238,11 +240,17 @@
                         ['view', ['codeview']]
                     ];
                 } else {
+                    if (styletags) {
```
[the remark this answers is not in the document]
gagnieray: You're right. Thanks for pointing out that mistake. I'll fix that!

galette/templates/default/page.html.twig
gagnieray: The plugin I'm working on needs to be able to "control" code set in the javascript block of the pages templates. But I couldn't just move the "elements/scripts.html.twig" after the javascript block in the templates as this could break the code defined in that block.

That's why I ended with this solution.

FYI, one goal of my module is to provide a Consent Management Platform.

gagnieray:
> > [...] And the default configuration of summernote editor may not be suited to a plugin's needs.
> 
> Hum... I guess you have that need so I'm OK.

Thanks !

Indeed I need that. My plugin aims to provide some legal details to the user. Admins and staff could edit up to 3 pages to that intent, that should be enough to fullfill the requirements of different international legislations (legal information, terms and conditions, privacy policy).

Thus, I don't think that editors should be able to add images in the content of those pages. Nor be able to use `h1` as public pages already define one.

galette/templates/default/page.html.twig
trasher: My idea was to move the javascript block and your new code at the end of the `scripts.html.twig` file; not only your new code (sorry if I was not clear enough).

gagnieray: As `scripts.html.twig` is an include, moving the javascripts block this way would break inheritance in templates that extend `page.html.twig` and `public_page.html.twig` :confused:

trasher: Oh.. OK, I did not think this would break. OK for me then!

trasher:
Can you please add issues on the tracker so this can be part of next milestone changes? Thanks!

gagnieray:
> Can you please add issues on the tracker so this can be part of next milestone changes? Thanks!

Of course. I'll do that!

PR #615, opened 2025-03-04
Make galette:twig-cache comand usable for plugins

trasher:
With this change, you can use the following in the plugin `Makefile`:

```
twig :
        pushd ../../../../ && bin/console galette:twig-cache $(shell basename $(abspath $(CURDIR)/..))
```

This is a lot of `.` and `/` but it should work :rofl:

### DIFF
--- a/galette/lib/Galette/Console/Command/MakeTwigCache.php
+++ b/galette/lib/Galette/Console/Command/MakeTwigCache.php
@@ -27,6 +27,7 @@ use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Exception\InvalidOptionException;
 use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Twig\Environment;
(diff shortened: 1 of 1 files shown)


trasher:
I probably have to improve the cache dir part, it will fail if it does not exists, but it should also create it if it does not exists...

gagnieray:
> ```
> twig :
>         pushd ../../../../ && bin/console galette:twig-cache $(shell basename $(abspath $(CURDIR)/..))
> ```

Because `make` uses `/bin/sh`, `pushd` cannot be used in a Makefile.

Changing `INSTALLDIR = .` to `INSTALLDIR = $(shell pwd)` and then using the following rule is working :

```
twig :
         pushd ../../../../ && bin/console galette:twig-cache $(shell basename $(abspath $(CURDIR)/..))
```

trasher:
> > ```
> > twig :
> >         pushd ../../../../ && bin/console galette:twig-cache $(shell basename $(abspath $(CURDIR)/..))
> > ```
> 
> Because `make` uses `/bin/sh`, `pushd` cannot be used in a Makefile.
> 
> Changing `INSTALLDIR = .` to `INSTALLDIR = $(shell pwd)` and then using the following rule is working :
> 
> ```
> twig :
>          cd ../../../../ && bin/console galette:twig-cache $(shell basename $(abspath $(CURDIR)/..)) && cd ${INSTALLDIR}
> ```

Both works on my side, maybe because I use zsh? Anyway, I see no problem to change `INSTALLDIR`, will do that on all official plugins.

gagnieray:
> Both works on my side, maybe because I use zsh?

I use zsh too. With Debian as OS. This may explain the different behaviour.

trasher:
> > Both works on my side, maybe because I use zsh?
> 
> I use zsh too. With Debian as OS. This may explain the different behaviour.

Indeed yes. Anyway, your proposal is OK for me ;) I'll push PRs on plugins soon.

trasher:
See usage in https://github.com/galette/plugin-auto/pull/28

PR #616, opened 2025-03-06
Improve groups display in galleries

gagnieray:
fixes #1916

### DIFF
--- a/galette/templates/default/pages/public/gallery.html.twig
+++ b/galette/templates/default/pages/public/gallery.html.twig
@@ -47,22 +47,30 @@
     {% set mgroups = member.getManagedGroups() %}
     {% if member.isStaff() or mgroups|length %}
         <div class="extra content">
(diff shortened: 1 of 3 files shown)


trasher:
Perfect for tooltips, thank you :)

A small remark; maybe both status and group manager labels should be left aligned, current display seems a bit strange to me:
image

What's your opinion?

trasher:
(Note: this is visible on my side when logged in. When no user is logged, entries are less large, so this is not visible)

trasher:
> [...]
> What's your opinion?

Also, maybe could we use a different way to display status and groups management, like FUI label maybe?

I let you see, you often have better ideas than mine on the UI point of view ;)

trasher:
> [...]
> Are you OK with that ?

Perfect :ok_hand:

trasher:
Manually merged :)

PR #623, opened 2025-03-15
Rework dynamic translations

trasher:
The very first goal orf this PR is to fix https://bugs.galette.eu/issues/1894; by adding a checksum column to be used in SQL primary key rather than relying on original text (cause issues on key length on mysql since utfbmb4 migration...).

Secondary goal is to rework a bit the page itself.

### DIFF
--- a/galette/includes/routes/management.routes.php
+++ b/galette/includes/routes/management.routes.php
@@ -325,6 +325,11 @@ $app->post(
 )->setName('doRemoveStatus')->add($authenticate);
 
 $app->get(
+    '/dynamic-translation/{text_orig_sum}',
+    [DynamicTranslationsController::class, 'dynamicTranslation']
+)->setName('dynamicTranslation')->add($authenticate);
+
+$app->get(
     '/dynamic-translations[/{text_orig}]',
     [DynamicTranslationsController::class, 'dynamicTranslations']
 )->setName('dynamicTranslations')->add($authenticate);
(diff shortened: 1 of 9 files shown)


trasher:
Regarding UI rework.

With actual proposition, adding a new string (seems just better than the existing):
image

Updating an existing one:
image

Removing extra "useless" accordion entry makes UI better I think. I'm not really convinced with dropdown display, but I guess this can be fixed.
The main problem is I've tried to replace the original "onsubmit" with simple links; but the result is not really correct. I don't know if I've choose the right direction.

@gagnieray could you please give me your opinion? Thank you :)

gagnieray:
> The main problem is I've tried to replace the original "onsubmit" with simple links; but the result is not really correct. I don't know if I've choose the right direction.
> 
> @gagnieray could you please give me your opinion? Thank you :)

Yes you took the right direction :+1: 

I would just use a "basic" styled [Search In-Menu](https://fomantic-ui.com/modules/dropdown.html#search-in-menu) instead of a "secondary" styled Search Dropdown button.

Have a look at my coming commit to see the difference.

trasher:
> I would just use a "basic" styled [Search In-Menu](https://fomantic-ui.com/modules/dropdown.html#search-in-menu) instead of a "secondary" styled Search Dropdown button.
> 
> Have a look at my coming commit to see the difference.

Yes, that's perfect; thank you :+1:

PR #624, opened 2025-03-19
Add complementary replacement values from the preferences

gagnieray:
At first, I had the intention to add some of these values in the plugin I'm working on. But I realized that emails and pdf models would not be able to use them :confused: 

Then I tought I would open a request to add a feature to allow plugins to create replacement values available to the Core :thinking: 

But I found the following existing (old) requests on the tracker :
- [#201](https://bugs.galette.eu/issues/201) : Téléphone siège social de l'association
- [#204](https://bugs.galette.eu/issues/204) : siren, siret, N° greffe association

So, I hope it will be OK to add them in Core. This PR adds the following fields in the "General" tab of the Preferences :
- telephone number (with the possibility to select a staff member's one)
- email address (different from the sender email address)
- registration number
- registration office
- business identification number

I also add JS code to show/hide the staff member select when relevant.

settings

I chose not to add dedicated fields for the french SIREN *and* SIRET numbers, because unless I'm wrong, other countries don't make this distinction. In France, and because Galette doesn't deal with multiple addresses ; only the SIREN *or* a SIRET is necessary (not both). Moreover, nothing prevents the users from setting both number in the corresponding field (a.k.a "business identification number").

I didn't deal with the other requests of the issues above : 

- "heures de permanence" (201 & 204) : I don't see a convenient way to implement that. Maybe a simple textarea for "Additionnal contact details" could be enough? Let me know if you want me to add this textarea ; 201 could then be closed after that.

- "carte d'identité administrative de l'association" (204) : it can already exist on the website of the association, or could be implemented in a plugin. Moreover, the same thing is asked in another related issue but as a homepage ([#494](https://bugs.galette.eu/issues/494)).

- "adresse siège social" / "adresse de gestion" (204) : that would mean multiple addresses, and possibly each with it's own phone number, email (and SIRET for france). I don't think it is the purpose of Galette to list all the existing addresses of an association.

- "références Journal Officiel déclaration" (204) : this one is totally useless in my opinion

- "page statuts" (204) : statuses can be published as a document ; or I could add it in my plugin if requested

- "références bancaires BIC et IBAN..." (204) : they could be easily be added. Let me know if you want me to add them.

### DIFF
--- a/galette/lib/Galette/Core/Preferences.php
+++ b/galette/lib/Galette/Core/Preferences.php
@@ -53,6 +53,10 @@ use Galette\Repository\Members;
  * @property string $pref_pays Country
  * @property integer $pref_postal_address Postal address to use, one of self::POSTAL_ADDRESS*
  * @property integer $pref_postal_staff_member Staff member ID from which retrieve postal address
+ * @property string $pref_org_phone_number Phone number
+ * @property integer $pref_org_phone Phone number to use, one of self::PHONE_NUMBER*
+ * @property integer $pref_org_phone_staff_member Staff member ID from which retrieve phone number
+ * @property string $pref_org_email Email address
  * @property boolean $pref_disable_members_socials Disable social networks for members
  * @property string $pref_lang Default instance language
  * @property integer $pref_numrows Default number of rows in lists
@@ -175,6 +179,13 @@ class Preferences
     /** Postal address will be the one of the selected staff member */
     public const POSTAL_ADDRESS_FROM_STAFF = 1;
 
+    /** Phone number will be the one given in the preferences */
+    public const PHONE_NUMBER_FROM_PREFS = 0;
+    /** Phone number will be the one of the selected staff member */
+    public const PHONE_NUMBER_FROM_STAFF = 1;
+    /** Phone number will be the GSM of the selected staff member */
+    public const PHONE_NUMBER_MOBILE_FROM_STAFF = 2;
+
     /** Public pages stuff */
     /** Public pages are publicly visibles */
     public const PUBLIC_PAGES_VISIBILITY_PUBLIC = 0;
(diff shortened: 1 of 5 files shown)


trasher:
Well... A few remarks.

I'm not sure all "administrative numbers" (SIREN/SIRET, RNA, DUNS, APE, and so on) are *really* useful in replacements for Galette. I'm pretty sure "registration office" is quite useless.
Adding few more fields also make UI and preferences management more complex :/

So, even if it could be logical those information being available as fields in preferences related to the association, it begins to look like a need for dynamic fields on settings side (that would be way more powerful, and permit to ensure nothing will be forget). 

Issues were opened on tracker issues years ago by one user; I do not remember anyone else has been asking for that since. Most of those issues are as is way to complex or out of the Galette scope.

> Then I tought I would open a request to add a feature to allow plugins to create replacement values available to the Core 🤔

Yes, it could be interesting, no existing plugin had information to add there. It may not be simple to achieve, I never really think about that.

>     * "heures de permanence" (201 & 204) : I don't see a convenient way to implement that. Maybe a simple textarea for "Additionnal contact details" could be enough? Let me know if you want me to add this textarea ; 201 could then be closed after that.

I'm not going to add (nor accept) that (see next point).

>     * "carte d'identité administrative de l'association" (204) : it can already exist on the website of the association, or could be implemented in a plugin. Moreover, the same thing is asked in another related issue but as a homepage ([#494](https://bugs.galette.eu/issues/494)).

My general answer is Galette is not a CMS. [A request to add a "landing page"](https://vote.galette.eu/posts/82/page-interne-association-vie-du-club) has also been done.
IMO, most of Galette users have a separate website to present their activities; which certainly can include a full contact page with all required details.

>     * "adresse siège social" / "adresse de gestion" (204) : that would mean multiple addresses, and possibly each with it's own phone number, email (and SIRET for france). I don't think it is the purpose of Galette to list all the existing addresses of an association.

No indeed. For structures with several geographic entities; there are many other problems that would be addressed. For that kind of case, using one Galette instance per entity is certainly the best option (but Galette is probably not a real option at all).

>     * "références Journal Officiel déclaration" (204) : this one is totally useless in my opinion

Useless, yes.

>     * "page statuts" (204) : statuses can be published as a document ; or I could add it in my plugin if requested

Yes, it can be done as a document.

>     * "références bancaires BIC et IBAN..." (204) : they could be easily be added. Let me know if you want me to add them.

It can be proposed as a document as well.

Also, several of those information can also be added directly in PDF models, or mail signature.

trasher:
> So, does that mean you are OK just for the telephone number and the email address ? 😁

Yes, those ones really makes sense ;-)

trasher (approved):

PR #628, opened 2025-03-22
Use FUI radio inputs and checkboxes; fixes #1922

gagnieray:
With also few other slight improvements at the same time: 
- fix leading dots in plugins init db modal
- narrow some columns in lists tables (the first and/or last ones ; usually for selections and actions)


trasher (approved):
I did not see any problem :)

PR #629, opened 2025-03-24
More improvements on dynamic translations page

gagnieray:
The page UI can still be improved a little I think :smile: 

I chose to disable the original text input because I think that it should not be modified, and I added a button to delete translation if needed (translations of preferences cannot be removed automatically when a value is modified).

Other changes are listed in the comment of the commit.

modify

modal

### DIFF
--- a/galette/includes/routes/management.routes.php
+++ b/galette/includes/routes/management.routes.php
@@ -340,6 +340,11 @@ $app->post(
 )->setName('editDynamicTranslation')->add($authenticate);
 
 $app->get(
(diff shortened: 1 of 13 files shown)


gagnieray:
The more I digg into dynamic translations, the more I discover issues :sweat_smile:

trasher:
> The more I digg into dynamic translations, the more I discover issues 😅

:rofl: 

Yes, "dynamic" things are quite complex to handle, and existing code is not always as correct as it should... As far as I remember, all of that never fully worked as it should have :(

PR #630, opened 2025-03-27
More improvements on dynamic translations page and some others fixes

gagnieray:


galette/templates/default/pages/configuration_payment_types.html.twig
```diff
@@ -73,7 +73,7 @@
                                     <span class="ui special popup">{{ _T("Edit '%s' payment type")|replace({'%s': ptype.getName()}) }}</span>
                                 </a>
                                 <a
-                                    href="{{ url_for('dynamicTranslations', {'text_orig': ptype.getName(false)|escape}) }}"
+                                    href="{{ url_for('dynamicTranslations', {'text_orig': ptype.getName(false)}) }}"
```
trasher, line 76: Are you sure we can remove that escape (and other related to dynamic translation route? As far as I remember, escaping here was mandatory to prevent some characters to break the route (like `&` and so on).

trasher (commented):
I did not yet have time to test, but that seems correct, expect my interrogation on escaping.

galette/templates/default/pages/configuration_payment_types.html.twig
gagnieray: OK... :thinking: 

As far as I have tested, it appears that the characters that can break routes are `/` and `\` (ends with a 404 error).

`#` seem to be the character added by escaping and that is related to [#1927](https://bugs.galette.eu/issues/1927) :confused: 

`?` is another character that can cause the same problem...

I will look into other escaping twig possibilities.

trasher: I wonder if twig capacities will allow such specific replacements. An alternative would be to keep escape as is; but to handle it differently on controller/entity class side.

Since the value stored in database is correct, this should not affect existing instances. I think I'll have a few hours to work on Galette this week-end - I could take a deeper look at al of this.

gagnieray: > I wonder if twig capacities will allow such specific replacements. An alternative would be to keep escape as is; but to handle it differently on controller/entity class side. [...]

`|escape('url')` might be a better candidate as it (almost) solves the problem. With it, the only problematic character remaining seems to be `/` (which would be logical as in the documentation of twig it is stated that "This should not be used to escape an entire URI; only a subcomponent being inserted").

gagnieray: I think I found the solution : call the route using simply the query parameters (already handled on the controller side) instead of using the route placeholder. This way, escaping in twig seems not even necessary.

trasher:
Merged as part of #634

PR #631, opened 2025-04-05
Use documents types translated names in lists; fixes #1930

gagnieray:


trasher:
Unless I missed something, with this proposal, calling `Document::getType()` will return translated version of the type if it's a system one; and original text version if it's a user one.
Also, this method is used to display data (so it shoudl rely on translated values) but also to store data in database (chere it must use the original version).

I did not have time to look closer at the issue, but after a quick look at the code, I guess a solution would be to call translation on `getType()` call result template side; or to add a new method which always returns a translated text.

gagnieray:
> Unless I missed something, with this proposal, calling `Document::getType()` will return translated version of the type if it's a system one; and original text version if it's a user one. 

Exactly. 

In the actual state of code, the types added by a user are not translatable ; the values are stored as is in the database, and then displayed on the frontend. 

Only the system's types use keys defined as constants with corresponding full names and translations. The keys are stored in the database when a document is added, and therefore they are also the values displayed on the frontend.


> Also, this method is used to display data (so it shoudl rely on translated values) but also to store data in database (chere it must use the original version).

Yes. `Document::getType()` is used on the template side to display the types in the management list.

For the public page, it is `Document::getTypedList()` (which calls `Document::getType()`) that returns the types with the corresponding documents.


> I did not have time to look closer at the issue, but after a quick look at the code, I guess a solution would be to call translation on `getType()` call result template side; or to add a new method which always returns a translated text.

Of course, take time to look closer to my changes :wink: 

They do not, in principle, change the values stored in the database. They only change the value returned for the system types and displayed on the frontend.

Tell me if this was the right way to do it, or if I shall indeed add a new method to do this.

gagnieray:
> In the actual state of code, the types added by a user are not translatable

Actually they are... I didn't pay attention at first...

gagnieray:
So... My changes are obviously not the solution :sweat_smile:

trasher:
Not tested, but something like this may be enought:
```diff
diff --git a/galette/templates/default/pages/documents_list.html.twig b/galette/templates/default/pages/documents_list.html.twig                                                                                                              
index c600069ee..e0f8a8d49 100644                                                                                                                                                                                                             
--- a/galette/templates/default/pages/documents_list.html.twig                                                                                                                                                                                
+++ b/galette/templates/default/pages/documents_list.html.twig                                                                                                                                                                                
@@ -72,7 +72,7 @@                                                                                                                                                                                                                             
                     {{ document.getDocumentFilename() }}                                                                                                                                                                                     
                 </a>                                                                                                                                                                                                                         
             </td>
-            <td data-col-label="{{ _T("Type") }}">{{ document.getType() }}</td>
+            <td data-col-label="{{ _T("Type") }}">{{ _T(document.getType()) }}</td>
             <td data-col-label="{{ _T("Visibility") }}">{{ document.getPermissionName() }}</td>
             <td data-col-label="{{ _T("Creation date") }}">{{ document.getCreationDate() }}</td>
             {% if mode != 'ajax' %}
```

gagnieray:
Yes. That fixes things on the management list :+1: 

Then, there is only one other thing to fix : the translations of the system's types in the select of the form to add a document. In this select, translations used for the system's types are the ones defined in core and not the ones eventually customized using the labels translations.

trasher (approved):
Seems correct; thank you!

gagnieray:
> This one is still draft, is there a reason, or should I merge?

I was still looking to resolve the case of the system's type names in the select when adding a new document : when a dynamic translation exists for a system's type, it is always the name defined in `Document::getSystemTypes()` that is used and not the dynamic translation.

It's a bit confusing when the system's types available in the select do not have the same translations as the ones displayed in the lists. 

But I think this can not be resolved without reworking dynamic translations. It is another issue or limitation of the feature ([#1928](https://bugs.galette.eu/issues/1928)).

So, you can merge :wink:

PR #633, opened 2025-04-11
Paid contributions; closes #1859

trasher:


trasher:
Not a real blocker, but I'm not fan of boolean field display in that case:
image

As far as I know, there is no way to display boolean components fields as "block" like all other; that would be a solution.  I though about alternatives, but none was really better.

@gagnieray maybe would you see another way to display that new field?

gagnieray:
> @gagnieray maybe would you see another way to display that new field?

I'll think about that :wink:

gagnieray:
Why not simply move the field at the end of the form, after the comments field ?

Otherwise, I noticed some issues : 
- as the field is required, it is not possible to change the status when editing a contribution
- when modifying a contribution attached to a transaction that is not totally dispatched, the user is redirected to the membership fee creation form (I'm not sure it is really an issue, because, the amount is automatically set to the undispatched amount of the transaction. It makes sense when adding contributions to a transaction, but it is a bit disturbing when simply editing an existing one in such case)

Finally, I think a few things are missing : 
- a corresponding column in the contributions list
- a replacement value that could be used in email and pdf contents
- batch actions to set a selection of contributions as paid or unpaid

trasher:
> Why not simply move the field at the end of the form, after the comments field ?

Hum... I'll take a look, but I'm not really convinced.

>     * as the field is required, it is not possible to change the status when editing a contribution

Yes, IO'm not sure it makes sense to make a boolean field required; I saw the mistake when posting the screenshot; I'll fix that.

>     * when modifying a contribution attached to a transaction that is not totally dispatched, the user is redirected to the membership fee creation form (I'm not sure it is really an issue, because, the amount is automatically set to the undispatched amount of the transaction. It makes sense when adding contributions to a transaction, but it is a bit disturbing when simply editing an existing one in such case)

As far as I remember, this always has worked like that; that's not a bug ;)

> Finally, I think a few things are missing :
> 
>     * a corresponding column in the contributions list
> 
>     * a replacement value that could be used in email and pdf contents
> 
>     * batch actions to set a selection of contributions as paid or unpaid

Agree for all; I'll work on that.

Thanks for the feedback :+1:

trasher:
> > * a corresponding column in the contributions list

Added.

> > * a replacement value that could be used in email and pdf contents

Added.

> > * batch actions to set a selection of contributions as paid or unpaid

It's not possible to mass edit contributions; I won't work on that for now. If you think we should have mass edit on contribution, please open a enhancement ticket ;)

Paid is no longer required; and I did not change its place because it seems quite not logic having that elsewhere. As initially said, that's not a blocker; that's not that important.

PR #634, opened 2025-04-11
Optimize tables display

gagnieray:
**This PR is based on #630.**

The first commit added uses definition tables introduced in #630 to better display the first column of headers when viewing a member's details (as suggested by @Tezorc on the development mailing list).

The second commit added uses [FUI collapsing cells](https://fomantic-ui.com/collections/table.html#collapsing-cell) on a selection of specific columns (usually : ids, actions, dates, counts, amounts, ...). This way such columns only use as much space as required.

### DIFF
--- a/galette/lib/Galette/Controllers/DynamicTranslationsController.php
+++ b/galette/lib/Galette/Controllers/DynamicTranslationsController.php
@@ -73,15 +73,25 @@ class DynamicTranslationsController extends AbstractController
 
         try {
             $orig = $this->l10n->getStringsToTranslate();
+            $text_exists = false;
 
             if ($text_orig_sum === '') {
                 $text_orig_sum = array_key_first($orig);
             }
-            $text_orig = $orig[$text_orig_sum] ?? $text_orig_sum;
 
-            $params['exists'] = isset($orig[$text_orig_sum]);
+            if (isset($orig[$text_orig_sum]) || isset($orig[md5($text_orig_sum)])) {
+                $sum = isset($orig[$text_orig_sum]) ? $text_orig_sum : md5($text_orig_sum);
+                $text_exists = true;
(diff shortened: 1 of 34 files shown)


trasher:
Sounds globally OK to me, thanks :)

A little detail: TH text color on  definition table is maybe a too dark (seen on Labels translation page):
image

Also, not directly related but I saw that while checking some pages:
image

I do not know where this "yellow" is coming from - I'm pretty sure this is not intended :)

gagnieray:
> Sounds globally OK to me, thanks :)
> 
> A little detail: TH text color on definition table is maybe too dark (seen on Labels translation page):

Good catch! It seems that the default styles for this variation don't use the same color variable as standard tables.
I'll fix that.
 
> Also, not directly related but I saw that while checking some pages: image
> 
> I do not know where this "yellow" is coming from - I'm pretty sure this is not intended :)

I cannot reproduce on my side. 
This yellow color `#faffc3` doesn't even seem to exist in the theme :thinking:
Please show my the html code of a yellow row.

trasher:
> > Sounds globally OK to me, thanks :)
> > A little detail: TH text color on definition table is maybe too dark (seen on Labels translation page):
> 
> Good catch! It seems that the default styles for this variation don't use the same color variable as standard tables. I'll fix that.

Fixed, thank you :)

> [...]
> I cannot reproduce on my side. This yellow color `#faffc3` doesn't even seem to exist in the theme 🤔 Please show my the html code of a yellow row.

My bad, issue comes from the "stipe" plugin which overrides `.odd` background color with an `!important` instruction. Disabling this plugin solves the issue.

trasher (approved):

trasher:
I've pushed an extra commit that removes the use of `odd`/`even` classes; this was an old code.

PR #653, opened 2025-06-16
Add body parsing middleware

gagnieray:
Required by https://github.com/galette-community/plugin-stripe/pull/5

https://www.slimframework.com/docs/v4/middleware/body-parsing.html

### DIFF
--- a/galette/includes/main.inc.php
+++ b/galette/includes/main.inc.php
@@ -176,6 +176,9 @@ $app->add(function (Request $request, RequestHandler $handler) use ($container) 
     return $handler->handle($request);
 });
 
+// Add body parsing Middleware - required to parse posted json, form data and xml
+$app->addBodyParsingMiddleware();
+
 // Add Routing Middleware - required for ACLs to work
 $app->addRoutingMiddleware();
 


trasher:
At first view, this seems to have a low impact; but this will automatically impact request data, so I wonder if it can cause issues on Core or other plugins (that will be difficult to see). Also, Stripe plugin does not seems currently maintained.

I do not know at all how Stripe works, I only worked on Paypal (and it's been a while since I really dive into that). Isn't that possible to handle this plugin side? As far as I understood, not using the middleware would just require an extra `Galette::jsonDecode()` (or native `json_decode`) call, no?

gagnieray:
> At first view, this seems to have a low impact; but this will automatically impact request data, so I wonder if it can cause issues on Core or other plugins (that will be difficult to see). 

**As far as I understand**, it is not supposed to change anything regarding `POST` requests that use the `application/x-www-form-urlencoded` or `multipart/form-data` formats. 

In thoses cases `getParsedBody()` still returns the `POST` parameters the same way as usual in an array.

This middleware adds the possibility to use the same method for `JSON` and `XML` formats which are not supported in PSR-7 implementations. As stated in the link provided about the middleware : "It’s very common in web APIs to send data in JSON or XML format", and "As this is a common requirement, Slim 4 provides `BodyParsingMiddleware` to handle this task."

That's why I though it was the proper way to do it, and that it could benefit to other plugins that would use external APIs.


> Also, Stripe plugin does not seems currently maintained.

Damn... So you mean that nobody will answer to the PR I opened on the plugin's repo ? :confused: 

That's sad because I managed to make the plugin fully working with next Galette's version :sweat_smile:
 
In fact, even for Galette 0.9.4+ users, the plugin is broken since Stripe changed the way it sends requests to the webhook using `JSON` instead of `application/x-www-form-urlencoded`. Payments are still working, but the webhook and thus history and contributions creation are not.


> I do not know at all how Stripe works, I only worked on Paypal (and it's been a while since I really dive into that).

Actually, Stripe offers different ways to implement payments.

The way the original author of the plugin chose to implement it was using a custom payment form on Galette's side rather than relying on an external form as you did with Paypal.

The form on Galette's side uses a [JS API](https://docs.stripe.com/js).

Authentication to the service and payment use the [Payment Intents API](https://docs.stripe.com/api/payment_intents).

And then a [single webhook](https://docs.stripe.com/webhooks#only-listen-to-event-types-your-integration-requires) is required to handle events sent by Stripe and trigger actions in Galette once the payment has succeeded.

FYI, with the custom payment form, public donations were not possible with this plugin. It is now fixed with the changes I made. It was just a question of asking for the minimum required details (name and email) along the amount choice to unlogged users only.


> Isn't that possible to handle this plugin side? As far as I understood, not using the middleware would just require an extra `Galette::jsonDecode()` (or native `json_decode`) call, no?

Indeed, it must be possible to handle this on the plugin side decoding the request using `getBody()`.

Adding this middleware would however have the advantage that this task would be provided 'out of the box' by simply using `getParsedBody()` as usual and could benefit other plugins :grin: 

But I understand your concern to avoid any unwanted side effect !

I'll give a try to handle this on the plugin's side :wink:

PR #657, opened 2025-06-21
Add member filter on scheduled payments list like contributions

trasher:
closes #1945

### DIFF
--- a/galette/lib/Galette/Repository/ScheduledPayments.php
+++ b/galette/lib/Galette/Repository/ScheduledPayments.php
@@ -327,42 +327,7 @@ class ScheduledPayments
             $select->where->in('s.' . self::PK, $this->current_selection);
         }
 
-        if ($this->filters->member_filter != null) {
-            //handle case when list is filtered on a single member id
-            if (!$this->login->isAdmin() && !$this->login->isStaff() && $this->filters->member_filter != $this->login->id) {
-                $member = new Adherent(
-                    $this->zdb,
-                    (int)$this->filters->member_filter,
-                    [
-                        'picture' => false,
-                        'groups' => false,
-                        'dues' => false,
-                        'parent' => true
-                    ]
-                );
-                if (
-                    !$member->hasParent() ||
-                    $member->parent->id != $this->login->id
-                ) {
-                    //check if member is part of logged-in user managed groups
-                    $mgroup = $this->login->getManagedGroups();
-                    $groups = $member->getGroups();
-                    if (count(array_intersect(array_keys($mgroup), array_keys($groups))) == 0) {
-                        Analog::log(
-                            'Trying to display scheduled payments for member #' . $member->id .
-                            ' without appropriate ACLs',
-                            Analog::WARNING
-                        );
-                        $this->filters->member_filter = $this->login->id;
-                    }
-                }
-            }
-            $select->where(
-                [
-                    'a.' . Adherent::PK => $this->filters->member_filter
-                ]
-            );
-        }
+        $this->buildMemberWhereClause($select);
 
         try {
             if ($this->filters->start_date_filter != null) {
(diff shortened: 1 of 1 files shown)


galette/templates/default/pages/scheduledpayments_list.html.twig
```diff
@@ -194,7 +233,17 @@
...
+                {% if filters.member_filter == "" or filters.member_filter == null %}
+                    <a
+                            href="{{ url_for("scheduledPayments", {"option": "member", "value": mid}) }}"
+                            title="{{ _T("Show only '%name' transactions")|replace({"%name": mname}) }}"
+                    >
+                        <i class="filter grey small icon" aria-hidden="true"></i>
+                    </a>
+                {% endif %}
+                {{ memberName({id: scheduled.getContribution().member}) }}
+            </td>
```
gagnieray, line 246:
```suggestion
            {% if ((login.isAdmin() or login.isStaff()) and member is not defined) or pmember is defined %}
                <td data-col-label="{{ _T("Member") }}">
                    {% if filters.member_filter == "" or filters.member_filter == null %}
                        <a
                            href="{{ url_for("scheduledPayments", {"option": "member", "value": mid}) }}"
                            title="{{ _T("Show only '%name' scheduled payments")|replace({"%name": mname}) }}"
                        >
                            <i class="filter grey small icon" aria-hidden="true"></i>
                        </a>
                    {% endif %}
                     <a
                        href="{{ url_for("member", {"id": mid}) }}"
                        title="{{ _T("Show '%name' card")|replace({"%name": mname}) }}"
                    >
                        {{ memberName({id: scheduled.getContribution().member}) }}
                    </a>
                </td>
            {% endif %}
```

The title of the filter link should be related to scheduled payments.

Otherwise, like contributions, this column should be hidden when relevant (the same if statement must also be added on the column's header), and a link can be added on the member's name.

And I realized that, on a member's own list of scheduled payments, the member cannot view the payments of its children like with contributions.

Similary, group managers cannot have access to the scheduled payments list. The corresponding right is missing in the settings, and group managers cannot view the standard list of scheduled payments like with contributions.

trasher: OK, so it's more complex that what I had in mind, I'll postpone after 1.2.0 release

trasher: > Similary, group managers cannot have access to the scheduled payments list. The corresponding right is missing in the settings, and group managers cannot view the standard list of scheduled payments like with contributions.

I'm not going to open access to scheduled payments to groups managers. Managing access is already a nightmare - and already cause security issues.
You can open a ticket on the bug tracker for future reference.

gagnieray: > I'm not going to open access to scheduled payments to groups managers. Managing access is already a nightmare - and already cause security issues.

Actually it was not a proper request for that :wink: 

It was an observation : I'm wondering if it's necessary in both cases (parents and group managers). Maybe not.

To be clearer, what is only necessary in my opinion at this point to include in 1.2.0 is the member filtering with the suggestion code I made for hiding the member column when relevant.

> You can open a ticket on the bug tracker for future reference.

You mean a ticket about opening access to scheduled payments to groups managers ? So that your answer remains in the tracker and you can reject the request ?

trasher: > [...] 
> Actually it was not a proper request for that 😉
> 
> It was an observation : I'm wondering if it's necessary in both cases (parents and group managers). Maybe not.
> 
> To be clearer, what is only necessary in my opinion at this point to include in 1.2.0 is the member filtering with the suggestion code I made for hiding the member column when relevant.

Oh... OK, sorry, I kind of misread/misunderstood; trying to handle things too fast. I'll work on requested changes then ;)

> > You can open a ticket on the bug tracker for future reference.
> 
> You mean a ticket about opening access to scheduled payments to groups managers ? So that your answer remains in the tracker and you can reject the request ?

It would have been required for further investigation/reflexion on my side; but since that's just an observation, it's no need fr a ticket.

gagnieray (approved):
Reviewed and tested. Everything is working properly :+1: :rocket:

PR #660, opened 2025-06-26
Show due scheduled payments in list

gagnieray:


galette/lib/Galette/Entity/ScheduledPayment.php
```diff
@@ -464,6 +464,18 @@ public function setPaid(bool $is_paid = true): self
         return $this;
     }
 
+    /**
+     * Is payment due?
+     *
+     * @return boolean
+     */
+    public function isDue(): bool
```
trasher, line 472: Can you add a test for this new method please?

trasher (approved):

trasher:
Perfect, thank you :)

PR #663, opened 2025-06-28
Remove pref_show_id; closes #1949

trasher:


galette/templates/default/pages/preferences.html.twig
```diff
@@ -277,13 +277,6 @@
                             {% endfor %}
                         </select>
                     </div>
-                    <div class="field inline">
-                        <div class="ui right aligned toggle checkbox">
-                            <input type="checkbox" name="pref_show_id" id="pref_show_id" value="1" {% if pref.pref_show_id %} checked="checked"{% endif %}/>
-                            <label for="pref_show_id">{{ _T("Show identifiers") }}</label>
-                        </div>
-                        <i class="circular small basic question icon tooltip" title="{{ _T("Display database identifiers in related windows") }}" aria-hidden="true"></i>
-                    </div>
```
gagnieray, line 286: This would require to adjust the position of the field "Select a ratio" to balance the 2 colums. I will have a look at that as soon as possible this WE.

trasher: I'll merge this one anyway, there's no hurry (I'm OK with current display ;))

PR #666, opened 2025-07-06
Fix few invalid HTML

trasher:


galette/templates/default/elements/list.html.twig
```diff
@@ -115,7 +115,7 @@
                         </tfoot>
                     </table>
     {% endif %}
-
+                </div>
```
trasher, line 118: I have a doubt on this one; not sure it's correctly placed (I did not see any issues on display; but that does not mean it's OK :D

gagnieray: There's definitely a missing closing `div` tag.
But in my opinion, it shouldn't be added on line 118, but on line 138, before the closing `form` tag.

gagnieray: Or maybe simply removing the opening `div` tag on line 69 would be OK. A div here is useless I think.

trasher: Removing extra div seems OK, I'll keep this solution. I did not see any issues; it can be fixed in another PR if any ;)

PR #672, opened 2025-08-09
Fix display when no rights

trasher:
Ensure member not owned from manager does
not get displayed after error on
Contributions and Transactions add/edit


gagnieray:
Le Sat, 09 Aug 2025 00:24:07 -0700
Johan Cwiklinski ***@***.***> a écrit :

> @trasher requested your review on: galette/galette#672 Fix display
> when no rights.

Sorry, I won't be able to review your changes until monday :/

-- 

Guillaume AGNIERAY

gagnieray (commented):
> No problem, I still have some tests to fix apparently; have a good week-end ;)

Indeed. At the moment I'm facing the same error that is raised by the unit tests.

trasher:
Thanks for testing :)

Regarding the extra link, I tried to avoid it, but that would make things complicated.
I've fixed the submit button for transactions in my lateste commit

PR #673, opened 2025-08-11
Improve news system

trasher:
Improve News mechanism on dashboard, to allow several entries to be displayed (including from plugins).
* Galette news (from website) will be displayed only to Staff and Admin
* Asso news (RSS URL from Preferences - if different from the previous), will be used to be displayed to all users
* Plugin news will be displayed if present (plugins are in charge to handle ACLs)

### DIFF
--- a/galette/lib/Galette/Core/Preferences.php
+++ b/galette/lib/Galette/Core/Preferences.php
@@ -313,7 +313,7 @@ class Preferences
         /* New contribution script */
         'pref_new_contrib_script' => '',
         'pref_bool_wrap_mails' => true,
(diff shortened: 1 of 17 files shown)


galette/lib/Galette/Core/Preferences.php
```diff
@@ -313,7 +313,7 @@ class Preferences
         /* New contribution script */
         'pref_new_contrib_script' => '',
         'pref_bool_wrap_mails' => true,
-        'pref_rss_url' => 'https://galette.eu/dc/index.php/feed/atom',
+        'pref_rss_url' => Galette::RSS_URL,
```
trasher, line 316:
```suggestion
        'pref_rss_url' => '',
```

PR #688, opened 2025-09-09
Rework groups management page

trasher:
PoC

I do not really like the current groups management page. Here is an alternative proposal.

It of course needs some more work; this is a proposal to be discussed. ping @gagnieray ;)

### DIFF
--- a/galette/templates/default/elements/group.html.twig
+++ b/galette/templates/default/elements/group.html.twig
@@ -23,20 +23,6 @@
 {% set managers = group.getManagers() %}
 {% set members = group.getMembers() %}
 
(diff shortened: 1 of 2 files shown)


gagnieray:
In my opinion, it would be even better to use a list view with dragable items first (the groups' hierarchy could be managed from the list) : 

group_list

And then a form view, to add/edit all the details : 

group_form

But I suppose this UI would require more additional work in the controllers.

trasher:
OK, thanks for your feeback :) I keep this one open for now; I'll think about that once 1.2 will be released ;)

trasher:
Replaced with #692

PR #692, opened 2025-09-12
Draggable groups list

gagnieray:
In response to #688, here is my proposal for a draggable list view of groups :grin:

### DIFF
--- a/galette/lib/Galette/Controllers/Crud/GroupsController.php
+++ b/galette/lib/Galette/Controllers/Crud/GroupsController.php
@@ -79,7 +79,7 @@ class GroupsController extends CrudController
 
         return $response
             ->withStatus(301)
-            ->withHeader('Location', $this->routeparser->urlFor('groups', ['id' => (string)$id]));
+            ->withHeader('Location', $this->routeparser->urlFor('doEditGroup', ['id' => (string)$id]));
     }
 
 
@@ -126,7 +126,6 @@ class GroupsController extends CrudController
      * @param Response            $response PSR Response
      * @param string|null         $option   One of 'page' or 'order'
      * @param integer|string|null $value    Value of the option
-     * @param ?integer            $id       Member id to check rights
      *
      * @return Response
      */
@@ -134,8 +133,7 @@ class GroupsController extends CrudController
         Request $request,
         Response $response,
         ?string $option = null,
-        int|string|null $value = null,
-        ?int $id = null
+        int|string|null $value = null
     ): Response {
         $groups = new Groups($this->zdb, $this->login);
         $group = new Group();
@@ -144,40 +142,6 @@ class GroupsController extends CrudController
         $groups_root = $groups->getList(false);
         $groups_list = $groups->getList();
 
-        if ($id !== null) {
-            if ($this->login->isGroupManager($id)) {
-                $group->load($id);
-            } else {
-                Analog::log(
-                    'Trying to display group ' . $id . ' without appropriate permissions',
-                    Analog::INFO
-                );
-                return $response->withStatus(403);
-            }
-        }
-
-        if ($id === null && count($groups_list) > 0) {
-            $group = current($groups_list);
-            if (!$this->login->isGroupManager($id)) {
-                foreach ($groups_list as $g) {
-                    if ($this->login->isGroupManager($g->getId())) {
-                        $group = $g;
-                        break;
-                    }
-                }
-            }
-        }
-
-        $parent_groups = [];
-        foreach ($groups_list as $parent_group) {
-            if ($group->canSetParentGroup($parent_group)) {
-                $parent_groups[] = $parent_group;
-            }
-        }
-
-        //Active tab on page
-        $tab = $request->getQueryParams['tab'] ?? 'group_information';
-
         // display page
         $this->view->render(
             $response,
@@ -185,12 +149,67 @@ class GroupsController extends CrudController
             [
                 'page_title'            => _T("Groups"),
                 'groups_root'           => $groups_root,
-                'parent_groups'         => $parent_groups,
-                'group'                 => $group,
-                'tab'                   => $tab
+                'is_paginated'          => false,
+                'form'                  => true,
+                'table'                 => [
+                    'class' => 'unstackable',
+                    'tbody' => [
+                        'id'    => 'listed_groups',
+                        'class' => 'sortable-items'
+                    ]
+                ],
+                'nb'                    => count($groups_list)
             ]
         );
         return $response;
+    }
+
+    /**
+     * List reorder
+     *
+     * @param Request  $request  PSR Request
+     * @param Response $response PSR Response
+     *
+     * @return Response
+     */
+    public function reorderList(Request $request, Response $response): Response
+    {
+        $post = $request->getParsedBody();
+        $list = '<ul>';
+        if (isset($post['reordered']) && !empty($post['reordered'])) {
+            foreach ($post['reordered'] as $value) {
+                $item = explode('|', $value);
+                $id = $item[0];
+                $parentId = $item[1];
+                $group = new Group((int)$id);
+                $parentGroup = new Group((int)$parentId);
+                if ($parentId != '0') {
+                    $group->setParentGroup((int)$parentId);
+                    $change = str_replace(
+                        '%1',
+                        $parentGroup->getName(),
+                        _T("new parent is %1")
+                    );
+                } else {
+                    $group->detach();
+                    $change = _T("parent removed");
+                }
+                $group->store();
+                $list .= '<li>' . $group->getName() . ' (' . $change . ')</li>';
+            }
+            $list .= '</ul>';
+            $this->flash->addMessage(
+                'success_detected',
+                str_replace(
+                    '%1',
+                    $list,
+                    _T("The parent of the following groups has been successfully modified : %1")
+                )
+            );
+        }
+        return $response
+            ->withStatus(301)
+            ->withHeader('Location', $this->routeparser->urlFor('groups'));
     }
 
     /**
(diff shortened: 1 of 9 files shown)


galette/templates/default/pages/groups_list.html.twig
```diff
@@ -18,364 +18,292 @@
...
+                {{ _T("All groups PDF") }}
+            </a>
         {% endif %}
         {% if login.isAdmin() or login.isStaff() %}
-                    <a href="{{ url_for("add_group", {"name": "NAME"}) }}" class="ui labeled icon fluid button tooltip" id="newgroup">
-                        <i class="plus circle green icon" aria-hidden="true"></i>
-                        {{ _T("New group") }}
-                    </a>
+            <a href="{{ url_for("add_group", {"name": "NAME"}) }}" class="ui tiny labeled icon button tooltip" id="newgroup">
+                <i class="plus green icon" aria-hidden="true"></i>
```
trasher: As used on other pages:
```suggestion
                <i class="plus circle green icon" aria-hidden="true"></i>
```

gagnieray:
This PR is completed I think :smile: 

Feel free to test and review the code when 1.2.0 is released ; there is no hurry !

In the meantime, here are some screenshots of the result : 

* List when user is admin or staff
  1-list_admin

* List when user is group manager with only edit permission
  2-list_manager_edit

* List when user is group manager with export permission but without edit permission
  3-list_manager_noedit_export

* Group edit form when user is admin or staff
  4-group_edit_admin

* Group view when user is group manager without edit permission
  5-group_manager_noedit

trasher:
Looks great; I'll test ASAP. Once again, thank for your work :+1:

galette/templates/default/pages/groups_list.html.twig
```diff
@@ -18,364 +18,292 @@
...
+    {{ parent() }}
+{% endblock %}
+
+{% block extra_header %}
+    <tr>
+        <th class="blue " colspan="3" style="padding-top:2px;padding-bottom:4px;" scope="row">
+            <span class="ui small text"><i class="info icon"></i> {{ _T("Move an item at the top of the list to reset its parent") }}</span>
+        </th>
+    </tr>
+{% endblock %}
```
trasher, line 131: I wonder if this message is really needed. That seems pretty standard; and almost same as previous behavior, no?

trasher (commented):
Seems OK from my tests; I really prefer this UI to the previous :+1:

gagnieray:
> Seems OK from my tests; I really prefer this UI to the previous 👍

I'm glad you like it :grin: 

But I have to admit I'm not entirely satisfied with the drag-and-drop functionality for groupmanagers. As it stands, unless completely removing drag-and-drop for groupmanagers, I haven't found a better solution than preventing saving in the event of unallowed movements.

galette/templates/default/pages/groups_list.html.twig
gagnieray: Well, it seems pretty obvious. But only if you drag a sub group at the top of the list at first. 

Some might not understand how to remove a parent if they only try to move a group between to others at the root of the tree.

And I fear that some might think that the drag-and-drop only opens the possibility to manually change the order groups in the list.

trasher:
> > Seems OK from my tests; I really prefer this UI to the previous 👍
> 
> I'm glad you like it 😁
> 
> But I have to admit I'm not entirely satisfied with the drag-and-drop functionality for groupmanagers. As it stands, unless completely removing drag-and-drop for groupmanagers, I haven't found a better solution than preventing saving in the event of unallowed movements.

Maybe drag&drop can be completely removed to make things simplier?

gagnieray:
> > But I have to admit I'm not entirely satisfied with the drag-and-drop functionality for groupmanagers. As it stands, unless completely removing drag-and-drop for groupmanagers, I haven't found a better solution than preventing saving in the event of unallowed movements.
> 
> Maybe drag&drop can be completely removed to make things simplier?

You decide ! :smile: 

Tell me, and I can remove the feature if you think it's better that way :wink:

trasher:
> Tell me, and I can remove the feature if you think it's better that way 😉

If d&d is removed, everything is OK for both of us, right?

In that case, yeah, go drop it, it's still possible to set parent group from edit form, and it's less error prone.

gagnieray:
> > Tell me, and I can remove the feature if you think it's better that way 😉
> 
> If d&d is removed, everything is OK for both of us, right?
> 
> In that case, yeah, go drop it, it's still possible to set parent group from edit form, and it's less error prone.

Alright ! I'll do that ASAP :+1:

gagnieray:
It seems you missed the last forced push I did 2 months ago (79d8a4a) :confused: 

Several of my last changes and code refactoring are not in your commit c4ba8c5
Including the d&d behaviour I mentioned yesterday.
So you may change your mind if you haven't really test what I was talking about :sweat_smile: 

I rebased the very last version of my code.
What should I do ? A forced push on this branch ? Or should I open a new PR ?

trasher:
Ha... Sorry for the mistake :/

You can of course force push current PR, no problem. Changes I made were just CS related, nothing important.

I'll take another look asap; and I'lI keep you in touch.

galette/lib/Galette/Controllers/Crud/GroupsController.php
```diff
@@ -144,55 +142,76 @@ public function list(
...
-            if ($this->login->isGroupManager($id)) {
-                $group->load($id);
-            } else {
-                Analog::log(
-                    'Trying to display group ' . $id . ' without appropriate permissions',
-                    Analog::INFO
-                );
-                return $response->withStatus(403);
-            }
-        }
```
trasher, line 157: I guess this security check should be kept

galette/lib/Galette/Controllers/Crud/GroupsController.php
```diff
@@ -318,7 +337,39 @@ public function filter(Request $request, Response $response): Response
...
+        $group = new Group();
+        $group->setLogin($this->login);
+
+        $groups_list = $groups->getList();
+
+        if ($this->login->isGroupManager($id)) {
+            $group->load($id);
+        } else {
+            Analog::log(
+                'Trying to display group ' . $id . ' without appropriate permissions',
```
trasher, line 350:
```suggestion
                'Trying to edit group ' . $id . ' without appropriate permissions',
```

galette/lib/Galette/Entity/Group.php
```diff
@@ -248,14 +248,6 @@ private function loadSubGroups(): void
...
             $select = $zdb->select(self::TABLE, 'a');
 
-            if (!$this->login->isAdmin() && !$this->login->isStaff()) {
-                $select->join(
-                    ['b' => PREFIX_DB . self::GROUPSMANAGERS_TABLE],
-                    'a.' . self::PK . '=b.' . self::PK,
-                    []
-                )->where(['b.' . Adherent::PK => $this->login->id]);
-            }
-
```
trasher, line 258: Are we sure this should be removed?

trasher:
I did not test yet, just made a very quick code review.

Tests failures are not related to your changes; I'll fix them on develop branch.

galette/lib/Galette/Controllers/Crud/GroupsController.php
gagnieray: It is not necessary anymore in the `list` function as it is not possible anymore to display a group details on the list page. This check is now used in the `edit` function (see your comment https://github.com/galette/galette/pull/692#discussion_r2484867364).

galette/lib/Galette/Controllers/Crud/GroupsController.php
gagnieray: Maybe it should be "edit or display" as it depends on the permissions setting ?

galette/lib/Galette/Entity/Group.php
gagnieray: Currently, on the `develop` branch, this select seems to be "faulty" as all the subgroups managed by a groupmanager that are not subgroups of a parent group also managed by the groupmanager can't be displayed in the tree.

So I choose to completely remove it in this PR and display the full tree to groupmanagers. Maybe this select can be fixed to properly display all the groups actually managed, but I wasn't able to do it.

trasher: OK, I see. That souds legit :)

galette/lib/Galette/Controllers/Crud/GroupsController.php
trasher: OK, fine :)

galette/lib/Galette/Controllers/Crud/GroupsController.php
trasher: That does not really matters, let's keep the original. Main point is unexpected attempts should be logged

trasher:
I've take time to test, I'm still not convinced with drag&drop :) I'd vote to remove it; but that's only my point of view.

If you want to keep that behavior, I guess you should add a message when something has changed to inform user it must be validated (just like when a member is added in group/managers).

trasher:
> Tests failures are not related to your changes; I'll fix them on develop branch.

This is fixed on develop ;)

gagnieray:
> I've take time to test, I'm still not convinced with drag&drop :) I'd vote to remove it; but that's only my point of view.

Do you mean completely ? Even for admins and staff ? Or only for groupmanagers ?

trasher:
> > I've take time to test, I'm still not convinced with drag&drop :) I'd vote to remove it; but that's only my point of view.
> 
> Do you mean completely ? Even for admins and staff ? Or only for groupmanagers ?

UI must not be different per profile, so either we keep for everyone, or we remove for everyone ;)

gagnieray:
> > > I've take time to test, I'm still not convinced with drag&drop :) I'd vote to remove it; but that's only my point of view.
> > 
> > 
> > Do you mean completely ? Even for admins and staff ? Or only for groupmanagers ?
> 
> UI must not be different per profile, so either we keep for everyone, or we remove for everyone ;)

Alright. Then my question is : in your opinion, is the drag and drop feature a relevant UI to modify groups' parents ? 🤔

If not, then I will completely drop this feature.

trasher:
Not, that's not irrelevant :)

OK, so. Just keep current behavior; we can adjust later after user feedback (if there is nay).

So, the only thing to do is to display a message like this one:
image

gagnieray:
> Not, that's not irrelevant :)
> 
> OK, so. Just keep current behavior; we can adjust later after user feedback (if there is nay).
> 
> So, the only thing to do is to display a message like this one: image

Alright, let's do that :+1:

trasher (approved):

trasher:
Once again, thanks for your job :) I've rebased, waiting for test to complete, and I'll merge.

PR #698, opened 2025-09-18
Contribution form template improvements 

gagnieray:
fixes #1960

### DIFF
--- a/galette/lib/Galette/Entity/ContributionsTypes.php
+++ b/galette/lib/Galette/Entity/ContributionsTypes.php
@@ -236,7 +236,7 @@ class ContributionsTypes
 
         try {
             $select = $this->zdb->select(self::TABLE);
-            $fields = [self::PK, 'libelle_type_cotis', 'amount'];
+            $fields = [self::PK, 'libelle_type_cotis', 'amount', 'cotis_extension'];
             $select->quantifier('DISTINCT');
             $select->columns($fields);
             $select->order(self::PK);
@@ -252,7 +252,8 @@ class ContributionsTypes
             foreach ($results as $r) {
                 $list[$r->{self::PK}] = [
                     'label' => _T($r->libelle_type_cotis),
-                    'amount' => $r->amount
+                    'amount' => $r->amount,
+                    'extension' => $r->cotis_extension
                 ];
             }
             return $list;
--- a/galette/templates/default/pages/contribution_form.html.twig
+++ b/galette/templates/default/pages/contribution_form.html.twig
@@ -372,58 +372,85 @@
     {% endif %}
 
     $(function() {
-    {% if not contribution.id -%}
-        var _types_amounts = {
-            {%- for key, values in type_cotis_options -%}
-               {%- if values.amount >= 0 and values.amount != '' -%}
-                    {{ key }}: {{ values.amount }},
-                {%- elseif values.amount == '' -%}
-                    {{ key }}: '',
-                {%- endif -%}
-            {%- endfor -%}
-        };
-        $('#id_type_cotis').dropdown({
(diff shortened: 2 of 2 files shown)


trasher:
It works well with membership extension as far as I can see.
With a beginning of membership date, it does not (but proposed fix does not seem to affect that at all -- and I'm not even sure how this should work :() That therefore seems another issue.

Regarding UI changes, could you resume what have changed please?

trasher:
Seems not correct on contributions list. On current develop branch:
image

With your changes:
image

I'm not against improvements on that part, but please do it in a separate PR we can merge after 1.2.0 release :)

gagnieray:
> I'm not against improvements on that part, but please do it in a separate PR we can merge after 1.2.0 release :)

Done !

FYI, on the actual PR, I added a glow transition effect on the fields when their values are updated on a contribution type change (amount and extension).

gagnieray:
> With a beginning of membership date, it does not (but proposed fix does not seem to affect that at all -- and I'm not even sure how this should work :() That therefore seems another issue.

Do you mean that with a beginning of membership date, the "end date of membership" on the form should also be update according to the duration defined in the settings of a contribution type (for exemple 4 months) ?

Then the other improvements for which I opened a separate PR are totally irrelevant, and that's indeed another issue.

trasher:
> > I'm not against improvements on that part, but please do it in a separate PR we can merge after 1.2.0 release :)
> 
> Done !

:+1: thank you!

> 
> FYI, on the actual PR, I added a glow transition effect on the fields when their values are updated on a contribution type change (amount and extension).

OK, no problem. I'll review this one quickly.

> > With a beginning of membership date, it does not (but proposed fix does not seem to affect that at all -- and I'm not even sure how this should work :() That therefore seems another issue.
> 
> Do you mean that with a beginning of membership date, the "end date of membership" on the form should also be update according to the duration defined in the settings of a contribution type (for exemple 4 months) ?

I guess it should, yes... It can cause issues with "offered months", but that's not been designed to work with a monthly membership.

Currently, if you first choose a monthly contribution, and then you change the selected member, the end date is changed - and that seem the correct way to go.
I'll open another issue on the tracker; no one yet complains about that problem. I even do not know if associations which uses a begin date of membership really uses monthly contributions types.

gagnieray:
> > Do you mean that with a beginning of membership date, the "end date of membership" on the form should also be update according to the duration defined in the settings of a contribution type (for exemple 4 months) ?
> 
> I guess it should, yes... It can cause issues with "offered months", but that's not been designed to work with a monthly membership.
> 
> Currently, if you first choose a monthly contribution, and then you change the selected member, the end date is changed - and that seem the correct way to go.

OK ! That makes sense :+1: 

> I'll open another issue on the tracker; no one yet complains about that problem. 

Fine !

> I even do not know if associations which uses a begin date of membership really uses monthly contributions types.

Indeed. Personnaly, I wrongly presumed that using a beginning of membership date always concerned yearly subscription. And that was the intent for the other PR #700 : to hide monthly choices in that configuration :sweat_smile:

trasher:
> Indeed. Personnaly, I wrongly presumed that using a beginning of membership date always concerned yearly subscription. And that was the intent for the other PR #700 : to hide monthly choices in that configuration 😅

Well, maybe you're right; I can't say :/
I guess we should ask on users mailing list to have feedback from concerned users?

gagnieray:
> I guess we should ask on users mailing list to have feedback from concerned users?

I guess yes. But what if users that use a monthly configuration, even if there is only one, didn't subscribed to the mailing list, and ask for a revert afterwards :confused: ?

I'll get a deeper look at this point :wink:

trasher (approved):

PR #700, opened 2025-09-19
Contributions types settings improvements

gagnieray:
Simplify membership extension choices depending on the settings used.

### DIFF
--- a/galette/templates/default/pages/contribution_type_form.html.twig
+++ b/galette/templates/default/pages/contribution_type_form.html.twig
@@ -33,17 +33,24 @@
                 <label for="amount">{{ _T("Amount:") }}</label>
                 <input type="text" name="amount" id="amount" value="{{ entry.amount }}"/>
             </div>
-            {% set cotisextensions = {
-                (constant("Galette\\Entity\\ContributionsTypes::DEFAULT_TYPE")): _T("Default"),
-                (constant("Galette\\Entity\\ContributionsTypes::DONATION_TYPE")): _T("No"),
-                (1): _Tn('%1$s month', '%1$s months', 1)|format(1)
-            } %}
-            {% for m in 2..11 %}
-                {% set cotisextensions = cotisextensions + {(m): _Tn('%1$s month', '%1$s months', m)|format(m)} %}
-            {% endfor %}
+            {% if preferences.pref_membership_ext != "" %}
+                {% set cotisextensions = {
+                    (constant("Galette\\Entity\\ContributionsTypes::DEFAULT_TYPE")): _T("Default"),
+                    (constant("Galette\\Entity\\ContributionsTypes::DONATION_TYPE")): _T("No"),
+                    (1): _Tn('%1$s month', '%1$s months', 1)|format(1)
+                } %}
+                {% for m in 2..11 %}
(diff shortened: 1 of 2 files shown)


gagnieray:
Following your comments on #698 :

In its actual state, when "Beginning of membership" is used in the settings : 
- when you change the selected member, the end date is changed
- when you change the contribution type, the end date is not changed

So to make things "work" with monthly subscriptions, the end date should change too when you change the contribution type.

And basically, it means that, at first, the created subscription is always starting at the beginning date defined in the settings, and for the duration of the selected contribution type.

Then what should be the proper behaviour if the start date is changed ?

Shouldn't the end date be updated accordingly ?

gagnieray:
I did a test.

Beginning of membership : 01/01
Contribution type with extension of 4 months.
I manually changed the new membership fee start date : 01/05
I didn't change the end date, so according to the settings, the end date remains 30/04.

The resulted created contribution starts on 2025-05-01 and ends on 2025-04-30, with a duration of 0 :astonished: 

There is definitely an issue when "Beginning of membership" setting and monthly subscriptions are used...

At least, when the start date is changed on the form, the end date should be updated accordingly.

And contributions with a start date posterior to the end date should not be accepted on saving.

And so, this PR seams irrelevant.

gagnieray:
I'm closing this PR. See https://bugs.galette.eu/issues/1963

trasher:
Since I was myself confused with that buggy monthly mmbership feature; maybe this PR can be reopened; it should make things more clear.

galette/templates/default/pages/contribution_type_form.html.twig
```diff
@@ -33,17 +33,24 @@
...
+            {% else %}
+                {% set cotisextensions = {
+                    (constant("Galette\\Entity\\ContributionsTypes::DEFAULT_TYPE")): _T("Yes"),
+                    (constant("Galette\\Entity\\ContributionsTypes::DONATION_TYPE")): _T("No")
+                } %}
+            {% endif %}
             {% include "components/forms/select.html.twig" with {
                 id: 'cotis_extension',
-                value: entry.cotis_extension,
+                value: preferences.pref_membership_ext != "" ? entry.cotis_extension : entry.cotis_extension <= 0 ? entry.cotis_extension : '-1',
```
gagnieray, line 53: FYI, this is for the (very unlikely, but possible) case where a contribution type has already been configured with a monthly extension on existing instances.

galette/templates/default/pages/contributions_types_list.html.twig
```diff
@@ -91,11 +98,11 @@
                     </td>
                     <td data-col-label="{{ _T("Membership extension") }}">
     {% if entry.extra == -1 %}
-                        {{ _T("Default") }}
+                        {{ preferences.pref_membership_ext != "" ? _T("Default") : _T("Yes") }}
     {% elseif entry.extra == 0 %}
                         {{ _T("No") }}
     {% else %}
-                        {{ _Tn('%1$s month', '%1$s months', entry.extra)|format(entry.extra) }}
+                        {{ preferences.pref_membership_ext != "" ? _Tn('%1$s month', '%1$s months', entry.extra)|format(entry.extra) : _T("Yes") }}
```
gagnieray, line 105: Here too, this is for the (very unlikely, but possible) case where a contribution type has already been configured with a monthly extension on existing instances.

trasher (approved):
LGTM

PR #702, opened 2025-09-22
Required adaptations to test plugin routes

trasher:


galette/includes/routes/plugins.routes.php
```diff
@@ -105,7 +105,7 @@ function ($request, $response) use ($module, $container) {
                     )->setName($module['route'] . 'Info')->add($authenticate);
 
                     $f = $module['root'] . '/_routes.php';
-                    include_once $f;
+                    require $f;
```
trasher, line 108: I'm not specially fan of that kind of change... But with `_once`, plugin routing configuration is not correctly loaded when testing a plugin :/

trasher: Will be improved later; by removing the include. See https://bugs.galette.eu/issues/1964

PR #703, opened 2025-09-22
More contribution form template improvements

gagnieray:
Update end date of membership when contribution type or start date of membership is changed on new membership fee creation.

fixes #1963

### DIFF
--- a/galette/includes/core_acls.php
+++ b/galette/includes/core_acls.php
@@ -56,6 +56,7 @@ $core_acls = [
     '/(.*)?addContribution/i'           => 'groupmanager',
     '/(at|de)tach_contribution/i'       => 'groupmanager',
     '/contributionDates/i'              => 'groupmanager',
+    '/contributionExtension/i'          => 'groupmanager',
     '/(.+)?contribution(.+)?/i'         => 'staff',
     '/(.*)?addTransaction/i'            => 'groupmanager',
     '/(.*)?editTransaction/i'           => 'groupmanager',
--- a/galette/includes/routes/ajax.routes.php
+++ b/galette/includes/routes/ajax.routes.php
@@ -66,6 +66,11 @@ $app->group('/ajax', function (RouteCollectorProxy $app) use ($authenticate): vo
     )->setName('contributionDates')->add($authenticate);
 
     $app->post(
+        '/contribution/extension',
+        [AjaxController::class, 'contributionExtension']
+    )->setName('contributionExtension')->add($authenticate);
+
+    $app->post(
         '/contribution/members[/{page:\d+}[/{search}]]',
         [AjaxController::class, 'contributionMembers']
     )->setName('contributionMembers')->add($authenticate);
--- a/galette/lib/Galette/Controllers/AjaxController.php
+++ b/galette/lib/Galette/Controllers/AjaxController.php
@@ -297,7 +297,7 @@ class AjaxController extends AbstractController
     }
 
     /**
-     * Contributions dates
+     * Contribution dates
      *
      * @param Request  $request  PSR Request
      * @param Response $response PSR Response
@@ -327,7 +327,35 @@ class AjaxController extends AbstractController
     }
 
     /**
-     * Contributions dates
+     * Contribution extension
+     *
+     * @param Request  $request  PSR Request
+     * @param Response $response PSR Response
+     *
+     * @return Response
+     */
+    public function contributionExtension(Request $request, Response $response): Response
+    {
+        $post = $request->getParsedBody();
+
+        $contrib = new Contribution(
+            $this->zdb,
+            $this->login,
+            [
+                'type'  => (int)$post['fee_id']
+            ]
+        );
+
+        return $this->withJson(
+            $response,
+            [
+                'membership_extension' => $contrib->extension
+            ]
+        );
+    }
+
+    /**
+     * Contribution members
      *
      * @param Request     $request  PSR Request
      * @param Response    $response PSR Response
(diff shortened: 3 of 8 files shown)


gagnieray:
To be honest, I'm still having difficulties to understand the logic to respect when "Beginning of membership" setting is used and the start date on the form is modified :sweat_smile:

When switching from a contribution type to another, I reused the ajax call already existing when changing the member on the form. Everything should be fine in this case.

But in the eventuality the start date is changed (which you said should never happen in real life), what should be the end date depending on the extension of the selected contribution type ? :thinking:

At the moment, what I did is that I define the new end date respecting precisely the extension of the contribution type selected, no matter what is the "Begining of membership" in the settings.

That means that in case where the begining of membership is set to 01/01 : if a contribution type selected has an extension of 6 months and the start date of membership in the form is changed to 2025-08-01, the end date will be updated to 2026-07-31.

Is it correct ? Or should the end date proposed be 2025-12-31 ?

trasher:
> When switching from a contribution type to another, I reused the ajax call already existing when changing the member on the form. Everything should be fine in this case.

:+1: 

> But in the eventuality the start date is changed (which you said should never happen in real life), what should be the end date depending on the extension of the selected contribution type ? 🤔

Contribution type extension is *not part of date computation* if global setting is not set to membership extension - see `Contribution::retrieveEndDate()` logic (https://github.com/galette/galette/blob/develop/galette/lib/Galette/Entity/Contribution.php#L287).
Those are 2 distinct modes of date computation that cannot be mixed. 

When using a begin of membership date, everyone is member at  the same time. If we set a begin of membership on 2025-01-01, end of membership will be 2025-12-31.
If you create a contribution today, dates will be the same.

I still think associations who uses this parameter probably won't change dates creating new contributions; but I keep it possible, so user can do what he wants (like entering a past contribution for example). There is no way to know what to do here, so user is in charge to fill both start and end dates.

gagnieray:
> Contribution type extension is _not part of date computation_ if global setting is not set to membership extension - see `Contribution::retrieveEndDate()` logic (https://github.com/galette/galette/blob/develop/galette/lib/Galette/Entity/Contribution.php#L287). Those are 2 distinct modes of date computation that cannot be mixed.

The logic you describe, and the part of the code that you are pointing at, only concern _contribution types_ that are set with default or no extension.

But for "monthly" _contributions types_ that have an extension greater than 0, the extension _is_ actually part of date computation and the mode chosen in the settings is not taken in account (https://github.com/galette/galette/blob/develop/galette/lib/Galette/Entity/Contribution.php#L284).


> When using a begin of membership date, everyone is member at the same time. If we set a begin of membership on 2025-01-01, end of membership will be 2025-12-31. If you create a contribution today, dates will be the same.

That make sense to me. And as I told you at first, I always have supposed that using the mode with a begin of membership date only concerned yearly memberships. But that's not the case when "monthly" _contributions types_ are configured. 

In the actual state of code, if you add a new membership fee today, first switching to a contribution with 6 months of extension, and then selecting the user, beginning of membership will be 2025-01-01, and the end of membership will be 2025-06-30 :confused: 

But given your explanation, that's not what is supposed to happen. 

And then : 

1. the statements in `Contribution::retrieveEndDate()` are faulty.
2. if not using membership extension in global settings, it should not be possible to choose an other extension that default or none when configuring _contributions types_.

I know that in real conditions, it's very unlikely that users will switch from a mode to another. 

But even if membership extension is not used in global settings at first, it is still possible to set monthly extensions when configuring _contribution types_ which are then used in computation. And that's what made things so confusing to me ! And that's also why I opened #700 at first, which was not so irrelevant finally :sweat_smile: 


> I still think associations who uses this parameter probably won't change dates creating new contributions; but I keep it possible, so user can do what he wants (like entering a past contribution for example). There is no way to know what to do here, so user is in charge to fill both start and end dates.

That makes sense. And then, in case users want to create past contributions, they will logically choose a start date corresponding to the beginning of membership in the settings. And the end date chosen should always logically be a year later.

trasher:
> [...]
> The logic you describe, and the part of the code that you are pointing at, only concern _contribution types_ that are set with default or no extension.

Indeed, it's a bug. I'll fix (and add tests).

>     1. the statements in `Contribution::retrieveEndDate()` are faulty.

Yes :)

>     2. if not using membership extension in global settings, it should not be possible to choose an other extension that default or none when configuring _contributions types_.

Yes.

> But even if membership extension is not used in global settings at first, it is still possible to set monthly extensions when configuring _contribution types_ which are then used in computation. And that's what made things so confusing to me ! And that's also why I opened #700 at first, which was not so irrelevant finally 😅

Indeed. 
Sorry, I was a bit confused myself on that part - I do not work much on that feature.
I guess the monthly membership stuff has been added for "membership extension" configuration only (see https://bugs.galette.eu/issues/1438). Probably my fault, I did not think about date of beginning setting.

Anyway; I guess it's more clear for both of us now :fireworks: 

I'll take an eye fix retrieve dates computation issue quickly.

gagnieray:
> Anyway; I guess it's more clear for both of us now 🎆

Yes, definitely ! :+1: 


> I'll take an eye fix retrieve dates computation issue quickly.

Fine ! I'll clean this PR and remove all the unnecessary changes :wink:

gagnieray:
> I'll clean this PR and remove all the unnecessary changes 😉

So. In the end this PR is irrelevant. 

As the contribution type extension is not supposed to be used in dates calculation, the dates don't need to be updated when the contribution type is changed.

And in the case a user changes the start date, it would be overcomplicated and unnecessary to automatically update the end date accordingly on the frontend side. Moreover, as you stated, users that would need to make such change would logically also change manually the end date.

Therefore, fcc86abb has already properly fixed #1963 on the backend side, and I'm closing this PR.

I'll open a new issue on the tracker regarding the other bug identified lately.

PR #707, opened 2025-09-26
Make sure login cannot be emptied; closes #1968

trasher:


trasher:
Existing tests are broken, this stil need some work :/

trasher:
Should be OK now :)

gagnieray:
> Should be OK now :)

It is, yes :+1:

PR #714, opened 2025-10-08
Logos

trasher:


galette/lib/Galette/Core/Logo.php
```diff
@@ -56,7 +57,30 @@ public function __construct()
...
     protected function getDefaultPicture(): void
     {
-        $this->file_path = realpath(_CURRENT_THEME_PATH . 'images/galette.png');
+        $now = new DateTime();
+        $special = ''; //default logo
+
+        // Halloween special logo
+        $compare_date = new DateTime(date('Y') . '-10-31');
+        $days_left = $compare_date->diff($now)->days;
+        if ($days_left <= 30) {
```
gagnieray: This way, won't the logo be displayed 30 days before AND after the date ?

galette/lib/Galette/Core/Logo.php
```diff
@@ -56,7 +57,30 @@ public function __construct()
...
+            $special = '_halloween';
+        }
+
+        // Xmas special logo
+        $compare_date = new DateTime(date('Y') . '-12-25');
+        $days_left = $compare_date->diff($now)->days;
+        if ($days_left <= 30) {
```
gagnieray: This way, won't the logo be displayed 30 days before AND after the date ?

gagnieray (commented):
This will mainly concern new installations and the demo, but in the case where users are not using their own logo, I wonder if the PNG file size should be optimized a bit more to avoid adding unnecessary weight to the pages. Or could we not use SVG instead?

galette/lib/Galette/Core/Logo.php
trasher: You're right, well catched!

trasher:
> This will mainly concern new installations and the demo, but in the case where users are not using their own logo, I wonder if the PNG file size should be optimized a bit more to avoid adding unnecessary weight to the pages. Or could we not use SVG instead?

I only have a SVG file for the original logo (~70Ko instead of ~250Ko) :(

Images are indeed quite huge (~250Ko, ~460Ko, ~850Ko). If I replace them with webp, each image will be ~150Ko for the same size.

Galette login page on develop branch is ~2,5Mo - 150Ko for the logo seems correct to me.

galette/lib/Galette/Core/Logo.php
```diff
@@ -56,7 +57,30 @@ public function __construct()
...
     protected function getDefaultPicture(): void
     {
-        $this->file_path = realpath(_CURRENT_THEME_PATH . 'images/galette.png');
+        $now = new DateTime();
+        $special = ''; //default logo
+
+        // Halloween special logo
+        $compare_date = new DateTime(date('Y') . '-10-31');
+        $date_diff = $compare_date->diff($now);
+        if ($date_diff->invert == 1 && $date_diff->days <= 30) {
```
gagnieray: The logo is not displayed the D day :confused:

gagnieray:
> If I replace them with webp, each image will be ~150Ko for the same size.

:+1: That's way better !

galette/lib/Galette/Core/Logo.php
trasher: Gasp! I did not think about that; I'll rework.

trasher: I've fixed that (I hope last :D) case!

gagnieray: Yes! It's OK now :+1:

trasher:
I finally missed something
image


:D :D :D :D

PR #723, opened 2025-10-16
Handle webp images transparency

trasher:
Fix https://bugs.galette.eu/issues/1976

* handle transparency on all file formats (in facts, only JPG does not comes with transparency capabilities, but that does not seem to cause any issue)
* refactor to prevent copy-pasted code 
* when logo is a WEBP, use a converted to PNG for use with TCPDF
* Fix wrong format and mime on default logo
* use PrintLogo for... ASSO_PRINT_LOGO replacement

### DIFF
--- a/galette/lib/Galette/Core/PrintLogo.php
+++ b/galette/lib/Galette/Core/PrintLogo.php
@@ -51,7 +51,29 @@ class PrintLogo extends Logo
         $this->file_path = $pic->getPath();
         $this->format = $pic->getFormat();
         $this->mime = $pic->getMime();
-        //anyways, we have no custom print logo
+        //anyway, we have no custom print logo
         $this->custom = false;
+    }
+
+    /**
+     * Returns current file full path
+     *
+     * @return string full file path
+     */
+    public function getPath(): string
+    {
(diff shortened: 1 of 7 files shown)


galette/lib/Galette/Core/PrintLogo.php
```diff
@@ -54,4 +54,24 @@ protected function getDefaultPicture(): void
...
+    {
+        if ($this->getFormat() !== 'webp') {
+            return $this->file_path;
+        }
+
+        //TCPDF does not support background transparency for WEBP images, create a PNG version
+        $converted_logo_path = sprintf('%s/%s.%s', GALETTE_CACHE_DIR, 'galette_printlogo_converted', 'png');
+        if (!file_exists($converted_logo_path)) {
+            $print_logo = imagecreatefromwebp($this->file_path);
+            imagepng($print_logo, $converted_logo_path);
```
trasher, line 75: This works as expected, but a getter should not copy files... A better solution would be to automatically convert WEBP PrintLogo to PNG at upload - since WEBP version will never be used anyway.

trasher: Not sure I really can do better

PR #735, opened 2025-11-09
Move post_contribution_test_ script to webroot

trasher:
Use Preferenices::getURL() instead of guess again
closes #1981

### DIFF
--- a/galette/webroot/post_contribution_test.php
+++ b/galette/webroot/post_contribution_test.php
@@ -1,14 +1,7 @@
-#!/usr/bin/php
 <?php
 
-/* vim: set expandtab tabstop=4 shiftwidth=4 softtabstop=4: */
-
 /**
- * Post configuration script test
- *
- * PHP version 5
- *
- * Copyright © 2013-2014 The Galette Team
+ * Copyright © 2003-2025 The Galette Team
  *
  * This file is part of Galette (https://galette.eu).
  *
@@ -24,18 +17,18 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with Galette. If not, see <http://www.gnu.org/licenses/>.
- *
- * @category  Core
- * @package   Galette
- *
- * @author Johan Cwiklinski <johan@x-tnd.be>
- * @copyright 2013-2014 The Galette Team
- * @license   http://www.gnu.org/licenses/gpl-3.0.html GPL License 3.0 or (at your option) any later version
- * @link      https://galette.eu
- * @since     Available since 0.7.5dev - 2013-06-10
  */
 
-$args = array();
+declare(strict_types=1);
+
+
+/**
+ * Post configuration script test
+ *
+ * @author Johan Cwiklinski <johan@x-tnd.be>
+ */
+
+$args = [];
 $internal = false;
 
 if (defined('STDIN')) {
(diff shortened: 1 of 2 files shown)


galette/webroot/post_contribution_test.php
```diff
@@ -1,14 +1,8 @@
 #!/usr/bin/php
```
gagnieray: Using the script as is with a Galette internal call throws a PHP error.

`PHP Fatal error:  strict_types declaration must be the very first statement in the script`

I suppose it's normal and only given as an example.
In real conditions, when using a Galette internal call, the shellbang is obviously not necessary, right ?

galette/webroot/post_contribution_test.php
gagnieray, line 51: It seems that `$_POST['galette_internal']` is missing.

Using a Galette internal call, I don't get the Analog message but :

```
127.0.0.1 - 2025-11-09 11:32:28 - DEBUG - Galette\IO\ExternalScript::send result: 1output:
Requested as HTTP POST with parameters:<br/>key: id | value: 856key: date | value: 2025-11-09key: type | value: membershipkey: amount | value: 50key: voucher | value: /var/www/galette/galette/data/cache/v1.2.0.1/pdf_contribs/contribution_856_facture.pdf
```

galette/webroot/post_contribution_test.php
```diff
@@ -68,13 +62,13 @@
...
+    //a successful script returns 0, we do not output anything
     $fp = fopen(__DIR__ . '/cache/galette_post_contrib_file.txt', 'w');
     fwrite($fp, $args);
     fclose($fp);
 } else {
-    $json_args = json_decode($args);
+    $json_args = json_decode((string) $args['params']);
     foreach ($json_args as $k => $v) {
-        echo 'key: ' . htmlspecialchars($k) . ' | value: ' . htmlspecialchars($v);
+        echo 'key: ' . htmlspecialchars((string) $k) . ' | value: ' . htmlspecialchars((string) $v);
```
gagnieray: I get this error in PHP logs : 

```
PHP Fatal error:  Uncaught Error: Object of class stdClass could not be converted to string in /var/www/galette/galette/webroot/post_contribution_test.php:71\nStack trace:\n#0 {main}\n  thrown in /var/www/galette/galette/webroot/post_contribution_test.php on line 72
```

gagnieray (approved):

galette/webroot/post_contribution_test.php
trasher: Yes, I took a look at that, and there is no "galette_internal" call. I did not find why it's been added, it's always been present. I prefer not to change that for now.

galette/webroot/post_contribution_test.php
trasher: Strange, I do not have this issue locally :/

galette/webroot/post_contribution_test.php
trasher: I do not reproduce on my side; but I was testing using a donation; maybe it appears on a new membership fee only?

trasher: OK, I was testing using the wrong branch... Last commit solve that issue.

galette/webroot/post_contribution_test.php
trasher: I was using the wrong branch... Indeed the shebang is not required, I've removed it.

gagnieray:
OK great ! Everything seems to be working as expected now. Thank you ! :+1:

gagnieray (approved):

PR #738, opened 2025-11-12
Post contribution script improvements

gagnieray:
* send the payment type id to the script (useful to test the payment type used)
* send a token to authenticate calls to the script and prevent forged requests

### DIFF
--- a/galette/lib/Galette/Entity/Contribution.php
+++ b/galette/lib/Galette/Entity/Contribution.php
@@ -1104,6 +1104,7 @@ class Contribution implements AccessManagementInterface
         global $preferences;
 
         $payment = [
+            'id'    => $this->getPaymentTypeId(),
             'type'  => $this->getPaymentType()
         ];
 
@@ -1202,6 +1203,21 @@ class Contribution implements AccessManagementInterface
         } else {
             return _T("Donation");
         }
+    }
+
+    /**
+     * Get payment type id
+     *
+     * @return int
+     */
+    public function getPaymentTypeId(): int
+    {
+        if ($this->payment_type === null) {
+            return 0;
+        }
+
+        $ptype = new PaymentType($this->zdb, $this->payment_type);
+        return $ptype->id;
     }
 
     /**
--- a/galette/lib/Galette/Entity/Contribution.php
+++ b/galette/lib/Galette/Entity/Contribution.php
@@ -1157,6 +1157,8 @@ class Contribution implements AccessManagementInterface
             $contrib['member'] = $member;
         }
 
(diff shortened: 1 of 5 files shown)


galette/lib/Galette/Entity/Contribution.php
```diff
@@ -1204,6 +1207,21 @@ public function getTypeLabel(): string
         }
     }
 
+    /**
+     * Get payment type id
+     *
+     * @return int
+     */
+    public function getPaymentTypeId(): int
```
trasher, line 1215: Could you please add a test on this new method?

galette/webroot/post_contribution_test.php
```diff
@@ -55,17 +58,31 @@
...
-    //we're called without arguments => exit.
-    die('No arguments.');
-}
-
 if (defined('STDIN')) {
-    //a successful script returns 0, we do not output anything
-    $fp = fopen(__DIR__ . '/cache/galette_post_contrib_file.txt', 'w');
+    if (empty($args)) {
+        //we're called without arguments => exit.
+        die('No arguments.');
```
trasher: I guess this can be kept a first like it was, since `emtpy([])` will return true.

gagnieray: > I guess this can be kept a first like it was, since `emtpy([])` will return true.

Are you sure ? Isn't `$args` receiving parameters from call on line 39 : `$args = stream_get_contents(STDIN);` ?

trasher: `$args` is set from either `stream_get_contents`, `$_POST` or `$_GET`. In each case, if we receive no arguments, script must end.
Doing it globally makes things easier to understand; you changes do not impact that as far as I can see.

galette/lib/Galette/Entity/Contribution.php
gagnieray: I don't understand why, but adding a test with the scheduled payment type fails :confused:
https://github.com/galette/galette/actions/runs/19293840673/job/55170780831

trasher: I'll take an eye doing the final review ;)

gagnieray: Alright ! Thanks :+1:

trasher: I've found the issue. Schedued payment need to be logged in; and `Contribution` object must be initialized once logged in, or this type was not available.

trasher:
I made a few changes; everything should be OK now. Let me know if that's OK for you, and we can merge :)

gagnieray:
> I made a few changes; everything should be OK now. Let me know if that's OK for you, and we can merge :)

It's all good! :smile: :rocket:

trasher (approved):

PR #739, opened 2025-11-24
Various CS fixes

trasher:
Some issues that has been found while working on a larger subject that cannot be finished now.


galette/lib/Galette/Core/Picture.php
```diff
@@ -821,8 +821,6 @@ private function resizeImage(string $source, string $ext, ?string $dest = null,
             }
             // Cropped image.
             $thumb_cropped = imagecreatetruecolor((int)$crop_width, (int)$crop_height);
-            // Cropped ratio.
-            $ratio = $crop_width / $crop_height;
```
trasher, line 825: This variable is not used.
@gagnieray can you please take an eye; it seems just useless there but I'd like to be sure I did not miss anything.

galette/lib/Galette/Core/Preferences.php
```diff
@@ -983,17 +982,16 @@ public function getPostalAddress(): string
     /**
      * Returns phone number
      *
-     * @return string phone number
+     * @return bool|int|string phone number
      */
-    public function getPhoneNumber(): string
+    public function getPhoneNumber(): string|bool|int
```
trasher: Not sure we really expect `bool` nor `int` here; to be checked.

galette/lib/Galette/Core/Preferences.php
```diff
@@ -1298,11 +1294,10 @@ public function __set(string $name, mixed $value): void
     /**
      * Get instance URL from configuration (if set) or guessed if not
      *
-     * @return string
+     * @return int|string|true
      */
-    public function getURL(): string
+    public function getURL(): int|bool|string
```
trasher: Not sure we really expect `bool` nor `int` here; to be checked.

galette/lib/Galette/Core/Picture.php
gagnieray: You're right it's useless and never used afterwards.

trasher: Thanks for looking :)

PR #740, opened 2025-12-03
Add types on class constants

trasher:


trasher:
Would require PHP 8.3

PR #748, opened 2025-12-18
Rework CI configuration

trasher:
Speedup composer and npm installations
Use GH actions cache fro composer and npm
Use GH actions cache for semantic builds (whiuch takes ~2 minutes)


.github/workflows/ci-linux.yml
```diff
@@ -95,22 +101,43 @@ jobs:
...
+          path: ~/.npm
+          key: "${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}"
           restore-keys: |
-            ${{ runner.os }}-galette-${{ matrix.php-version }}-
+            ${{ runner.os }}-npm-
+
+      - name: Semantic hash
+        id: semantic-hash
+        if: env.skip != 'true'
+        run: echo "semantichash=$(tar  cf - --sort=name ./ui --mtime='UTC 2025-01-01' | sha1sum | cut -d' ' -f1)" >> $GITHUB_OUTPUT
```
trasher:
```suggestion
        run: echo "semantichash=$(tar cf - --sort=name ./ui --mtime='UTC 2025-01-01' | sha1sum | cut -d' ' -f1)" >> $GITHUB_OUTPUT
```

PR #795, opened 2026-03-03
Fix gulp copy encoding on font files

trasher:


gulpfile.js
```diff
@@ -106,10 +106,11 @@ var paths = {
   extras: [
     {
       src: './node_modules/summernote/dist/font/*',
-      dest: 'webfonts/'
+      dest: 'webfonts/',
+      encoding: false
     }, {
       src: './node_modules/summernote/dist/lang/*.min.js',
-      dest: 'js/lang/'
+      dest: 'js/lang/',
```
trasher:
```suggestion
      dest: 'js/lang/'
```

gagnieray (commented):
I'm a bit late for the review, but I confirm that everything is OK with this change 😉

trasher:
Thanks for the feedback ;)

PR #796, opened 2026-03-04
Major compatibility version

trasher:
Following recent (and yet incoming) changes on plugins; the compat version must be bumped.
Switching to 1.3 for clarity (will also brings some new features).


trasher:
As draft for now, since this will require to move/rename upgrade scripts; but there are at least 2 existing PRs to merge before.

PR #798, opened 2026-03-10
Feature/plugin db version rework

trasher:


galette/lib/Galette/Core/Plugins.php
```diff
@@ -771,6 +791,38 @@ public function overridePrefs(string $id): void
...
+     * 1- Sets version to 'NULL'
+     *    => it won't be detected as an old version and won't be marked as "to update" until the plugin is really updated at least once.
+     *    => update won't be detected automatically, and since plugins DB updates are not so frequent, a plugin that would need an update in several years
+     *       won't be marked as "to update".
+     *
+     * 2- Sets version to plugin version,
+     *    => it will be set to current plugin version, there is no way to ensure that plugin is actually up to date
+     *    => update has not been reworked for now, but original version will be the wrong one for plugins that really need an update on this migration
+     *    => this seems this is the current way of doing things (nothing inform user plugin DB must be updated)
+     */
```
trasher, line 818: @gagnieray could you please tell me what's your opinion on that?
I'd vote for the 2nd solution; to me this is just the way it's working right now - but I'd appreciate another eyes than mine :)

gagnieray: From a user's point of view, I think it would be better if the update were automatically detected :thinking: 
So, I would vote for the second solution too :wink:

trasher: Thanks for your feedback. I'll go for that solution.

galette/lib/Galette/Controllers/PluginsController.php
```diff
@@ -161,6 +190,8 @@ public function initPluginDb(Request $request, Response $response, string $id):
             $install->atPreviousStep();
         } elseif (isset($post['install_prefs_ok'])) {
             $install->atEndStep();
+        } elseif(isset($post['install_dbwrite_ok'])) {
```
trasher:
```suggestion
        } elseif (isset($post['install_dbwrite_ok'])) {
```

PR #799, opened 2026-03-19
Ask for superadmin password to read current conf at install

trasher:
closes #1996

### DIFF
--- a/galette/install/steps/db.php
+++ b/galette/install/steps/db.php
@@ -88,22 +88,31 @@ if ($install->configurationFileExists() && $install->isUpgrade()) {
             <div id="install_dbconfig">
 <?php
 }
+if (!$install->isUpgrade()) {
 ?>
-        <div class="inline required field">
-            <label for="install_dbtype"><?php echo _T("Database type:"); ?></label>
-            <select name="install_dbtype" id="install_dbtype" class="ui dropdown nochosen"<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>>
-                <option value="mysql"<?php echo $install->getDbType() === GaletteDb::MYSQL ? ' selected="selected"' : ''; ?>>Mysql</option>
-                <option value="pgsql"<?php echo $install->getDbType() === GaletteDb::PGSQL ? ' selected="selected"' : ''; ?>>Postgresql</option>
-            </select>
-        </div>
+
+            <div class="inline required field">
+                <label for="install_dbtype"><?php echo _T("Database type:"); ?></label>
+                <select name="install_dbtype" id="install_dbtype" class="ui dropdown nochosen">
+                    <option value="mysql"<?php echo $install->getDbType() === GaletteDb::MYSQL ? ' selected="selected"' : ''; ?>>Mysql</option>
+                    <option value="pgsql"<?php echo $install->getDbType() === GaletteDb::PGSQL ? ' selected="selected"' : ''; ?>>Postgresql</option>
+                </select>
+            </div>
+<?php
+}
+?>
             <div class="inline required field">
                 <label for="install_dbhost"><?php echo _T("Host:"); ?></label>
-                <input type="text" name="install_dbhost" id="install_dbhost" value="<?php echo htmlspecialchars($_POST['install_dbhost'] ?? $install->getDbHost() ?? $install->isUpgrade() ? '' : 'localhost', ENT_QUOTES, 'UTF-8'); ?>" placeholder="localhost" required/>
+                <input type="text" name="install_dbhost" id="install_dbhost" value="<?php echo htmlspecialchars((($_POST['install_dbhost'] ?? '') || !$install->isUpgrade()) ? $install->getDbHost() : 'localhost', ENT_QUOTES, 'UTF-8'); ?>" placeholder="localhost" required/>
             </div>
+<?php if (!$install->isUpgrade()) { ?>
             <div class="inline required field">
                 <label for="install_dbport"><?php echo _T("Port:"); ?></label>
                 <input type="text" name="install_dbport" id="install_dbport" value="<?php echo (int)($_POST['install_dbport'] ?? $install->getDbPort() ?? $install->isUpgrade() ? '' : $default_dbport); ?>" placeholder="<?php echo $default_dbport; ?>" required/>
             </div>
+<?php
+}
+?>
             <div class="inline required field">
                 <label for="install_dbuser"><?php echo _T("User:"); ?></label>
                 <input type="text" name="install_dbuser" id="install_dbuser" value="<?php echo htmlspecialchars($_POST['install_dbuser'] ?? $install->getDbUser() ?? '', ENT_QUOTES, 'UTF-8'); ?>" required/>
@@ -116,10 +125,14 @@ if ($install->configurationFileExists() && $install->isUpgrade()) {
                 <label for="install_dbname"><?php echo _T("Database:"); ?></label>
                 <input type="text" name="install_dbname" id="install_dbname" value="<?php echo htmlspecialchars($_POST['install_dbname'] ?? $install->getDbName() ?? '', ENT_QUOTES, 'UTF-8'); ?>" required/>
             </div>
+<?php if (!$install->isUpgrade()) { ?>
             <div class="inline required field">
                 <label for="install_dbprefix"><?php echo _T("Table prefix:"); ?></label>
-                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>/>
+                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required/>
             </div>
+<?php
+}
+?>
         </div>
         </div>
         <div class="ui section divider"></div>
--- a/galette/install/steps/db.php
+++ b/galette/install/steps/db.php
@@ -57,11 +57,11 @@ if ($install->configurationFileExists()) {
     ?>
         <div class="grouped fields">
             <div class="ui compact invisible checkbox existing_config">
-                <input type="radio" id="existing_config" name="config_choice">
+                <input type="radio" id="existing_config" name="config_choice" required="required">
                 <label for="existing_config" class="ui teal segment"><?php echo _T("Existing configuration file") ?></label>
             </div>
             <div class="ui compact invisible checkbox new_config">
-                <input type="radio" id="new_config" name="config_choice">
+                <input type="radio" id="new_config" name="config_choice" required="required">
                 <label for="new_config" class="ui teal segment "><?php echo _T("Enter your configuration") ?></label>
             </div>
         </div>
@@ -72,13 +72,13 @@ if ($install->configurationFileExists()) {
             <div class="field">
                 <div class="ui left icon input">
                     <i class="user icon" aria-hidden="true"></i><label for="login" class="visually-hidden"><?php echo _T("Username:"); ?></label>
-                    <input type="text" name="login" id="login" autofocus placeholder="<?php echo _T("Username:"); ?>"/>
+                    <input type="text" name="login" id="login" autofocus placeholder="<?php echo _T("Username:"); ?>" required="required"/>
                 </div>
             </div>
             <div class="field">
                 <div class="ui left icon input">
                     <i class="lock icon" aria-hidden="true"></i><label for="password" class="visually-hidden"><?php echo _T("Password:"); ?></label>
-                    <input type="password" name="password" id="password" placeholder="<?php echo _T("Password:"); ?>"/>
+                    <input type="password" name="password" id="password" placeholder="<?php echo _T("Password:"); ?>" required="required"/>
                 </div>
             </div>
             <input type="hidden" name="ident" value="1" />
--- a/galette/install/steps/db.php
+++ b/galette/install/steps/db.php
@@ -46,7 +46,7 @@ if ($install->isUpgrade()) {
 $default_dbport = $install->getDbType() === GaletteDb::PGSQL ? GaletteDb::PGSQL_DEFAULT_PORT : GaletteDb::MYSQL_DEFAULT_PORT;
 ?>
 <?php
-if ($install->configurationFileExists()) {
+if ($install->configurationFileExists() && $install->isUpgrade()) {
     ?>
         <div class="grouped fields">
             <div class="ui compact invisible checkbox existing_config">
@@ -79,7 +79,7 @@ if ($install->configurationFileExists()) {
     <?php
 }
 ?>
-        <div class="ui tab basic fitted segment<?php echo ($_POST['config_choice'] ?? '') === 'new_config' ? ' active' : ''; ?>" data-tab="new">
+        <div class="ui tab basic fitted segment<?php echo ($_POST['config_choice'] ?? '') === 'new_config' || !$install->isUpgrade() ? ' active' : ''; ?>" data-tab="new">
         <div class="inline required field">
             <label for="install_dbtype"><?php echo _T("Database type:"); ?></label>
             <select name="install_dbtype" id="install_dbtype" class="ui dropdown nochosen"<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>>
--- a/galette/install/steps/db.php
+++ b/galette/install/steps/db.php
@@ -55,8 +55,16 @@ if ($install->isUpgrade()) {
 <?php
 if ($install->configurationFileExists()) {
     ?>
-        <div class="ui button existing"><?php echo _T("Existing configuration file") ?></div>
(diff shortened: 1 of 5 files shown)


trasher:
@gagnieray maybe could you help a bit on this one on the UI part; if that's OK for you :)

See https://bugs.galette.eu/issues/1996

On the page that asks for DB information, we now should have 2 choices: either provide information, as usual; or enter superadmin credentials to use the existing config file.
I've tried with tabs component, but even if it's working, I'm not entirely convinced by the display - maybe you'll have better idea than mine, like usual ;p

I guess it currently fails because of a javascript line in common.js; related to the standard interface (not installer) - but since I was not happy of the result, I've not searched a lot yet.

gagnieray:
> @gagnieray maybe could you help a bit on this one on the UI part; if that's OK for you :)

Yes, I will gladly take a look at that :grin: 

Using tabs is the right solution in my opinion. However, instead of using a menu and displaying one of the tabs by default, I would prefer to [use buttons](https://fomantic-ui.com/modules/tab.html#without-menus) without displaying anything by default: the users make their choice by clicking a button; the corresponding form appears; that would seem more explicit to me.

See my upcoming commit :wink:

trasher:
Thanks!

 Usually this kind of displayed button in Galette has another meaning; users may not understand "it's a tab".

What would you think of using a horizontal accordion, like we already have in several places?

trasher:
Looking at this; I saw I need to know which config has been selected; and that's not possible with tabs or buttons (well, not without extra JS).

I've relied on checkbox (seehttps://fomantic-ui.com/modules/checkbox.html#invisible )
image

I've used the segment option, but this can be rendered as a button.
I'll go with that for now, so I can finalize the feature. Changing display will be quite easy, we can do that later.

gagnieray:
> I've used the segment option, but this can be rendered as a button.
> I'll go with that for now, so I can finalize the feature. Changing display will be quite easy, we can do that later.

:+1: 

But in the end, after looking into this, I'm wondering if there is a real interest in allowing users to reinstall Galette :thinking: 

In my opinion, it makes sense for the plugins to give users this choice. But I'm not sure it is the case for the core. At least, I don't think it is a common use case.

Shouldn't the whole process be simplified?: 
1. if no configuration file exists, only propose installing Galette.
2. if a configuration file exists, only propose updating.

trasher:
> > I've used the segment option, but this can be rendered as a button.
> > I'll go with that for now, so I can finalize the feature. Changing display will be quite easy, we can do that later.
> 
> 👍
> 
> But in the end, after looking into this, I'm wondering if there is a real interest in allowing users to reinstall Galette 🤔
> 
> In my opinion, it makes sense for the plugins to give users this choice. But I'm not sure it is the case for the core. At least, I don't think it is a common use case.
> 
> Shouldn't the whole process be simplified?:
> 
>     1. if no configuration file exists, only propose installing Galette.
> 
>     2. if a configuration file exists, only propose updating.

Hum... That's a good question.

First of all; I'm not sure that makes more sens for plugins than for core ^^ Also, remember PluginInstall code is entirely based on core install one (will be reworked a bit, see https://bugs.galette.eu/issues/1997, but currently it's not done at all).

For the 1, of course, and it's already done like that.
For the 2, that can make sense for developers and people who test Galette. Along with technical install rewrite, I'd like to simplify (an issue exists on tracker, number of steps should be reduced); and that would require a bigger UI rework (for example using ajax calls with loader, and just display a brief message if checks are OK, a modal if they are not. On my tests to reduce numbers of steps for plugins, I've also bypassed the query results page if everything is OK, but it could be great to have an alternative way to display executed queries anyway). 
Well, that's a bit out of the scope of this PR, and I guess I still need extra think about all of that (and open a ticket to list what must be done).

Requiring superadmin password on update still makes sens to me, as well as providing full database information as an alternative (note to myself: currently used to *update configuration*, that should be avoid; we should only compare values. Updating is not about changing host or port, that must be done directly in file.).

Current PR is related to update only; so it's still mergeable if OK :)

trasher:
> of this PR, and I guess I still need extra think about all of that (and open a ticket to list what must be done).

I've opened https://bugs.galette.eu/issues/2000 (the 2K one :D)

galette/install/steps/db.php
```diff
@@ -47,52 +40,89 @@
...
+                <input type="text" name="install_dbuser" id="install_dbuser" value="<?php echo $_POST['install_dbuser'] ?? $install->getDbUser(); ?>" required/>
             </div>
             <div class="inline required field">
                 <label for="install_dbpass"><?php echo _T("Password:"); ?></label>
                 <input type="password" name="install_dbpass" id="install_dbpass" value="" required/>
             </div>
             <div class="inline required field">
                 <label for="install_dbname"><?php echo _T("Database:"); ?></label>
-                <input type="text" name="install_dbname" id="install_dbname" value="<?php echo $install->getDbName(); ?>" required/>
+                <input type="text" name="install_dbname" id="install_dbname" value="<?php echo $_POST['install_dbname'] ?? $install->getDbName(); ?>" required/>
```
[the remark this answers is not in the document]
trasher: Fixed

galette/install/steps/db.php
```diff
@@ -47,52 +40,89 @@
...
             </div>
             <div class="inline required field">
                 <label for="install_dbport"><?php echo _T("Port:"); ?></label>
-                <input type="text" name="install_dbport" id="install_dbport" value="<?php echo $install->getDbPort() ?? $default_dbport; ?>" required/>
+                <input type="text" name="install_dbport" id="install_dbport" value="<?php echo $_POST['install_dbport'] ?? $install->getDbPort() ?? $install->isUpgrade() ? '' :$default_dbport; ?>" placeholder="<?php echo $default_dbport; ?>" required/>
             </div>
             <div class="inline required field">
                 <label for="install_dbuser"><?php echo _T("User:"); ?></label>
-                <input type="text" name="install_dbuser" id="install_dbuser" value="<?php echo $install->getDbUser(); ?>" required/>
+                <input type="text" name="install_dbuser" id="install_dbuser" value="<?php echo $_POST['install_dbuser'] ?? $install->getDbUser(); ?>" required/>
```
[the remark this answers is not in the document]
trasher: Fixed

gagnieray:
> I'm not sure that makes more sens for plugins than for core ^^

That's true :sweat_smile: 

> Also, remember PluginInstall code is entirely based on core install one (will be reworked a bit, see https://bugs.galette.eu/issues/1997, but currently it's not done at all).

:+1: 

> For the 1, of course, and it's already done like that. 

What I actually meant is that in step 2, during the initial install, the option to update should not be proposed.

> For the 2, that can make sense for developers and people who test Galette. 

Oh, OK. In my case, I just delete the configuration file; I don't mind having to refill all the required information about the database configuration (that's usual with all the other web applications I work with) :laughing: 

> [..] Requiring superadmin password on update still makes sens to me, as well as providing full database information as an alternative (note to myself: currently used to _update configuration_, that should be avoid; we should only compare values. Updating is not about changing host or port, that must be done directly in file.).

Well, for updates, Galette is the only web application I know that proposes that. All the others I use never ask for database information during an update; if any changes to the database are required, they usually have to be made manually in the configuration file.

> Current PR is related to update only; so it's still mergeable if OK :)

I must have missed something then. The title of the PR mentions "install". And (at least) at the moment the super admin password is requested whether a new install or an update is chosen in step 2.

trasher:
> > For the 2, that can make sense for developers and people who test Galette.
> 
> Oh, OK. In my case, I just delete the configuration file; I don't mind having to refill all the required information about the database configuration (that's usual with all the other web applications I work with) 😆

I have a different approach... In my configuration file, I have several working distincts setups that are commented but one; with mysql/postgres, for installation specific tests, etc). I usually have 4 or 5 of them; so I don't remove my configuration file :)

> 
> > [..] Requiring superadmin password on update still makes sens to me, as well as providing full database information as an alternative (note to myself: currently used to _update configuration_, that should be avoid; we should only compare values. Updating is not about changing host or port, that must be done directly in file.).
> 
> Well, for updates, Galette is the only web application I know that proposes that. All the others I use never ask for database information during an update; if any changes to the database are required, they usually have to be made manually in the configuration file.

Yep, I've changed that in one of the latest commits. Now, it should just check, on update, than existing parameters are the same than existing ones. So either you provide a superadmin credential, or you enter database information (a sysadmin may not know about Galette superadmin password).

> 
> > Current PR is related to update only; so it's still mergeable if OK :)
> 
> I must have missed something then. The title of the PR mentions "install". And (at least) at the moment the super admin password is requested whether a new install or an update is chosen in step 2.

A bug. It should indeed not be proposed at all at install :/

trasher:
> A bug. It should indeed not be proposed at all at install :/

Fixed :)

galette/install/steps/db.php
```diff
@@ -47,52 +38,89 @@
...
             </div>
             <div class="inline required field">
                 <label for="install_dbname"><?php echo _T("Database:"); ?></label>
-                <input type="text" name="install_dbname" id="install_dbname" value="<?php echo $install->getDbName(); ?>" required/>
+                <input type="text" name="install_dbname" id="install_dbname" value="<?php echo htmlspecialchars($_POST['install_dbname'] ?? $install->getDbName() ?? '', ENT_QUOTES, 'UTF-8'); ?>" required/>
             </div>
             <div class="inline required field info">
                 <label for="install_dbprefix"><?php echo _T("Table prefix:"); ?></label>
-                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required/>
+                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>/>
```
gagnieray: 2 things here : 
- the value used for update is not the actual value stored in the configuration file
- disabling the input prevents its value to be posted on submit

galette/install/steps/db.php
```diff
@@ -47,52 +38,89 @@
...
+            <input type="hidden" name="ident" value="1" />
+        </div>
+    <?php
 }
 ?>
+        <div class="ui tab basic fitted segment<?php echo ($_POST['config_choice'] ?? '') === 'new_config' || !$install->isUpgrade() ? ' active' : ''; ?>" data-tab="new">
         <div class="inline required field">
             <label for="install_dbtype"><?php echo _T("Database type:"); ?></label>
-            <select name="install_dbtype" id="install_dbtype" class="ui dropdown nochosen">
+            <select name="install_dbtype" id="install_dbtype" class="ui dropdown nochosen"<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>>
```
gagnieray: 2 things here :
- the option selected for update is not the actual value stored in the configuration file
- disabling the select prevents its value to be posted on submit

galette/install/steps/db.php
```diff
@@ -47,52 +38,89 @@
...
-                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required/>
+                <input type="text" name="install_dbprefix" id="install_dbprefix" value="<?php echo $install->getTablesPrefix() ?? 'galette_'; ?>" required<?php echo $install->isUpgrade() ? ' disabled="disabled"' : ''; ?>/>
 <?php
 if ($install->isUpgrade()) {
-    echo '<div class="ui compact floating orange message"><p>'
+    echo '<div class="ui orange label"><p>'
         . _T("(Indicate the CURRENT prefix of your Galette tables)")
         . '</p></div>';
 }
 ?>
```
gagnieray: If the intent is now to disable the table prefix field on update, this label can be removed.

gagnieray (changes requested):

galette/install/steps/db.php
```diff
@@ -47,52 +38,89 @@
...
+            <div class="field">
+                <div class="ui left icon input">
+                    <i class="user icon" aria-hidden="true"></i><label for="login" class="visually-hidden"><?php echo _T("Username:"); ?></label>
+                    <input type="text" name="login" id="login" autofocus placeholder="<?php echo _T("Username:"); ?>" required="required"/>
+                </div>
+            </div>
+            <div class="field">
+                <div class="ui left icon input">
+                    <i class="lock icon" aria-hidden="true"></i><label for="password" class="visually-hidden"><?php echo _T("Password:"); ?></label>
+                    <input type="password" name="password" id="password" placeholder="<?php echo _T("Password:"); ?>" required="required"/>
```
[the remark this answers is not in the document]
trasher: Handled by javascript

gagnieray:
Instead of checkboxes displayed as buttons, what would you think of a link (with js on the click event) in the message above the displayed form? This way a form could be displayed by default.

superadmin-credentials

database-credentials

trasher:
> Instead of checkboxes displayed as buttons, what would you think of a link (with js on the click event) in the message above the displayed form? This way a form could be displayed by default.

It seems less easy to read for users (short button text vs long link lost among others).
Another note: this is an informative message, it should not be displayed as a warning.

>  This way a form could be displayed by default.

Well, it could already; I just not have yet decided which one should appear by default ;)

gagnieray:
> It seems less easy to read for users (short button text vs long link lost among others).
> Another note: this is an informative message, it should not be displayed as a warning.

Another proposal : grouped buttons :grin: 

grouped-buttons

trasher:
Way better, yes!

Maybe not in grey? But OK for me, thanks :)

gagnieray:
With conditional buttons, the existing form is displayed by default.
And the button corresponding to the active form uses the "positive" color of FUI (green) : 

existing

new

trasher:
I was thinking about that, and I've an idea to propose.

What about this kind of display:
image

But with the forms in modals?

galette/lib/Galette/Core/Install.php
```diff
@@ -834,41 +833,78 @@ public function isGaletteInitStep(): bool
...
+    public function checkAgainstExistingConfig(array $values, array &$error_detected): bool
+    {
+        if ($this->configurationFileExists()) {
+            $existing = $this->loadExistingConfigFile([], true);
+            return $existing['db_host'] == $values['install_dbhost']
+                && $existing['db_port'] == $values['install_dbport']
+                && $existing['db_name'] == $values['install_dbname']
+                && $existing['db_user'] == $values['install_dbuser']
+                && $existing['db_pass'] == $values['install_dbpass']
+                && $existing['db_prefix'] == $values['install_dbprefix'];
```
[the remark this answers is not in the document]
trasher: Fixed; `db_` prefix is used everywhere now

galette/install/steps/db.php
trasher: Indeed... Since this form in case of an update is now just a check against the existing configuration, I guess host, port, user and pass are enough - so I'll remove display of prefix and type.

galette/install/steps/db.php
trasher: Same as install_dbprefix above

gagnieray:
> I was thinking about that, and I've an idea to propose.
> 
> What about this kind of display: [...] But with the forms in modals?

Well... I'm not convinced by the use of modals :thinking:

Usually, (correct me if I'm wrong) when a modal is used in Galette, the user remains on the view where the modal was originally opened.

Here, it would imply that when the chosen form is submitted in the modal, the user immediately goes to the next view/step. In my opinion, that would "break" the usual UX.

If the user chooses the "wrong" option, this would add an unnecessary click to close the modal and then select the other update option (unless the other option could be chosen in the opened modal). All this seems a bit cumbersome to me.

I believe the entire update process could be radically simplified without requiring any credentials at all.

I suppose that if you require credentials, it is for security reasons?

But, I might be wrong, the person who previously updated the files on the server is also the one supposed to handle the database upgrade when required, right?

So, once Galette is installed, why not just "lock" the installer? That's what is done by Dolibarr for example: before being able to launch a database upgrade, it is required to manually delete a lock file created on the server after the first install or the previous database upgrade. Therefore, the user can go through the entire update process without having to enter any database details.

What do you think?

trasher:
> Usually, (correct me if I'm wrong) when a modal is used in Galette, the user remains on the view where the modal was originally opened.
> 
> Here, it would imply that when the chosen form is submitted in the modal, the user immediately goes to the next view/step. In my opinion, that would "break" the usual UX.
> 
> If the user chooses the "wrong" option, this would add an unnecessary click to close the modal and then select the other update option (unless the other option could be chosen in the opened modal). All this seems a bit cumbersome to me.

Indeed, that's not a good idea.

> I believe the entire update process could be radically simplified without requiring any credentials at all.
> 
> I suppose that if you require credentials, it is for security reasons?

Exactly.

> But, I might be wrong, the person who previously updated the files on the server is also the one supposed to handle the database upgrade when required, right?

Honestly... No idea. I guess this will be the same person in most cases, yes.

> 
> So, once Galette is installed, why not just "lock" the installer? That's what is done by Dolibarr for example: before being able to launch a database upgrade, it is required to manually delete a lock file created on the server after the first install or the previous database upgrade. Therefore, the user can go through the entire update process without having to enter any database details.
> 
> What do you think?

Well... That would be secure, not so hard for users, and definitely require much less code. I guess I'll just go for that.

trasher:
Closing, see https://bugs.galette.eu/issues/2005

PR #800, opened 2026-03-19
"Hide" extra steps in plugin installation

trasher:
This permit to go from type selection to end at once


trasher:
Opened as reference, I do no like the way that works. See also https://bugs.galette.eu/issues/1997

PR #805, opened 2026-03-27
Hotfix/1.2.2

trasher:
I've added to much things; next release is not ready.
I've included all major fixes done in the develop branch.


trasher:
Finally, not a good idea. It's extra work for me; and I already have too much to do.

PR #826, opened 2026-04-15
Enable wcag21aa & RGAAv4 A11y checks

trasher:
Enable wcag21aa and rgaav4 in Playwright a11y tests.

A graphical report is available on build summaries:
https://github.com/galette/galette/actions/runs/24474947060?pr=826

Unfortunately, the full report with errors is too large to be displayed; the best option is probably to download the playwright report, unzip it and open it on a local instance with:
```
$ cd /path/to/galette
npx playwright show-report "/path/to/playwright-report"
```

For a complete Playwright guide, see https://github.com/galette/galette/blob/develop/README.tests.md#running-e2e-tests-playwright

### DIFF
--- a/galette/templates/default/components/form.html.twig
+++ b/galette/templates/default/components/form.html.twig
@@ -106,7 +106,6 @@
                         {% endif %}
                         {% if entry.field_id == 'email_adh' %}
                             {% set size = "30" %}
-                            {% set autocomplete = "new-password" %}
                         {% endif %}
                         {% if entry.field_id == 'fingerprint' %}
                             {% set size = "40" %}
--- a/galette/templates/default/components/form.html.twig
+++ b/galette/templates/default/components/form.html.twig
@@ -106,6 +106,7 @@
                         {% endif %}
                         {% if entry.field_id == 'email_adh' %}
                             {% set size = "30" %}
+                            {% set autocomplete = "new-password" %}
                         {% endif %}
                         {% if entry.field_id == 'fingerprint' %}
                             {% set size = "40" %}
--- a/galette/templates/default/elements/group.html.twig
+++ b/galette/templates/default/elements/group.html.twig
@@ -89,7 +89,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnmanagers_small">
+                    <a href="#" class="jsonly ui labeled icon button" id="btnmanagers_small">
                         <i class="user shield icon" aria-hidden="true"></i>
                         {{ _T("Manage managers") }}
                     </a>
@@ -109,7 +109,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnusers_small">
+                    <a href="#" class="jsonly ui labeled icon button" id="btnusers_small">
                         <i class="user icon" aria-hidden="true"></i>
                         {{ _T("Manage members") }}
                     </a>
--- a/galette/templates/default/elements/group.html.twig
+++ b/galette/templates/default/elements/group.html.twig
@@ -89,7 +89,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly ui labeled icon button" id="btnmanagers_small">
+                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnmanagers_small">
                         <i class="user shield icon" aria-hidden="true"></i>
                         {{ _T("Manage managers") }}
                     </a>
@@ -109,7 +109,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly ui labeled icon button" id="btnusers_small">
+                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnusers_small">
                         <i class="user icon" aria-hidden="true"></i>
                         {{ _T("Manage members") }}
                     </a>
--- a/galette/templates/default/elements/scripts.html.twig
+++ b/galette/templates/default/elements/scripts.html.twig
@@ -134,7 +134,7 @@
                 _compactMenu();
     {% endif %}
 
-    {% if autocomplete_options %}
+    {% if autocomplete is defined %}
                 $('#ville_adh_field, #lieu_naissance_field').search({
                     apiSettings: {
                         url: '{{ url_for('suggestTown', {'term': '{query}'}) }}',
--- a/galette/templates/default/elements/scripts.html.twig
+++ b/galette/templates/default/elements/scripts.html.twig
@@ -134,7 +134,7 @@
                 _compactMenu();
     {% endif %}
 
-    {% if autocomplete is defined %}
+    {% if autocomplete_options is defined %}
                 $('#ville_adh_field, #lieu_naissance_field').search({
                     apiSettings: {
                         url: '{{ url_for('suggestTown', {'term': '{query}'}) }}',
--- a/galette/templates/default/elements/scripts.html.twig
+++ b/galette/templates/default/elements/scripts.html.twig
@@ -134,7 +134,7 @@
                 _compactMenu();
     {% endif %}
 
-    {% if autocomplete %}
+    {% if autocomplete_options %}
                 $('#ville_adh_field, #lieu_naissance_field').search({
                     apiSettings: {
                         url: '{{ url_for('suggestTown', {'term': '{query}'}) }}',
--- a/galette/templates/default/pages/status_list.html.twig
+++ b/galette/templates/default/pages/status_list.html.twig
@@ -40,18 +40,18 @@
                 </thead>
                 <tfoot>
                 <tr>
-                    <td class="collapsing" data-scope="row">
+                    <td class="collapsing" data-scope="row" data-col-label="{{ _T("New status") }}">
                         <span class="visually-hidden">
                             {{ _T("New status") }}
                         </span>
                     </td>
-                    <td class="left" data-col-label="{{ _T("Name") }}">
+                    <td class="left">
                         <div class="required field">
                             <label for="libelle_statut">{{ _T("Name") }}</label>
                             <input size="40" type="text" name="libelle_statut" id="libelle_statut" required="required"/>
                         </div>
                     </td>
-                    <td class="left" data-col-label="{{ _T("Priority") }}">
+                    <td class="left">
                         <div class="required field">
                             <label for="priorite_statut">{{ _T("Priority") }}</label>
                             <input size="4" type="text" name="priorite_statut" id="priorite_statut" value="99" />
--- a/galette/templates/default/pages/status_list.html.twig
+++ b/galette/templates/default/pages/status_list.html.twig
@@ -39,7 +39,7 @@
                 </tr>
                 </thead>
                 <tfoot>
-                <tr>
+                <tr class="bottom aligned">
                     <td class="collapsing" data-scope="row" data-col-label="{{ _T("New status") }}">
                         <span class="visually-hidden">
                             {{ _T("New status") }}
--- a/galette/templates/default/pages/status_list.html.twig
+++ b/galette/templates/default/pages/status_list.html.twig
@@ -89,27 +89,24 @@
                         <td class="center actions_row collapsing">
                             <a
                                 href="{{ url_for('editStatus', {'id': eid}) }}"
-                                class="action single-edit"
+                                class="action single-edit icon-only"
                                 aria-label="{{ _T("Edit '%s' field")|replace({'%s': entry.name}) }}"
                             >
-                                <i class="ui edit icon tooltip" aria-hidden="true"></i>
-                                <span class="ui special popup">{{ _T("Edit '%s' field")|replace({'%s': entry.name}) }}</span>
+                                <i class="ui edit icon" aria-hidden="true"></i>
                             </a>
                             <a
                                 href="{{ url_for('dynamicTranslations', {}, {'text_orig': entry.text_orig}) }}"
-                                class="action single-translate"
+                                class="action single-translate icon-only"
                                 aria-label="{{ _T("Translate '%s'")|replace({'%s': entry.name}) }}"
                             >
-                                <i class="ui language grey icon tooltip" aria-hidden="true"></i>
-                                <span class="ui special popup">{{ _T("Translate '%s'")|replace({'%s': entry.name}) }}</span>
+                                <i class="ui language grey icon" aria-hidden="true"></i>
                             </a>
                             <a
                                 href="{{ url_for('removeStatus', {'id': eid}) }}"
-                                class="delete"
+                                class="delete icon-only"
                                 aria-label="{{ _T("Delete '%s' field")|replace({'%s': entry.name}) }}"
                             >
-                                <i class="ui trash red icon tooltip" aria-hidden="true"></i>
-                                <span class="ui special popup">{{ _T("Delete '%s' field")|replace({'%s': entry.name}) }}</span>
+                                <i class="ui trash red icon" aria-hidden="true"></i>
                             </a>
                         </td>
                     </tr>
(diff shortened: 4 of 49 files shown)


trasher:
@gagnieray seems like some of the reported issues are legitimate; for some others... Well, I'm not really sure I understand :/

gagnieray:
> [...] for some others... Well, I'm not really sure I understand :/

Which ones are you not sure to understand ?

Looking at the graphical reports, here are all the types of errors identified : 
- [label] Ensure every form element has a label
- [link-in-text-block] Ensure links are distinguished from surrounding text in a way that does not rely on color
- [link-name] Ensure links have discernible text
- [aria-input-field-name] Ensure every ARIA input field has an accessible name
- [aria-prohibited-attr] Ensure ARIA attributes are not prohibited for an element's role
- [autocomplete-valid] Ensure the autocomplete attribute is correct and suitable for the form field
- [color-contrast] Ensure the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds

Regarding the color contrasts, I have been working on it, and I will soon open a PR for that :wink: 

Playwright revealed some issues I hadn't noticed. However, I also fixed some that the tests didn’t reveal (obviously, it will be quite hard to test every case and get an exhaustive list of issues).

Otherwise, I noticed that the accessibility tests on the public members list and gallery are incorrect or incomplete. Playwrites shows the login page with the unauthorized message, not the actual public list or gallery :thinking:

trasher:
> > [...] for some others... Well, I'm not really sure I understand :/
> 
> Which ones are you not sure to understand ?

Well, I saw a problem about autocomplete, but on members list - but that was during development, I just took a quick look.

> Playwright revealed some issues I hadn't noticed. However, I also fixed some that the tests didn’t reveal (obviously, it will be quite hard to test every case and get an exhaustive list of issues).

Yes, of course; but having automated tests on that part can only improve Galette's accessibility :)

> Otherwise, I noticed that the accessibility tests on the public members list and gallery are incorrect or incomplete. Playwrites shows the login page with the unauthorized message, not the actual public list or gallery 🤔

Public pages tests are commented, aren't they?

By default, public pages are only allowed to logged in members. I've tried to change the setup; but had hard times dealing with selection in dropdowns.
But since display is almost the same, I can also use a member from fixture data to access them (and that, is something I can do). 

Not really easy to select a correct value in a dropdown - or I did not find yet the correct way to go. I'm quite new to Playwright - there are lots of things that are unknown to me ;)

trasher:
> Public pages tests are commented, aren't they?

Answering to myself: no, they aren't :D
I've commented on standard e2e tests, not on a11y... Will fix.

trasher:
I've changed the reporter, and ran a11y tests in a separate run.

Regarding remaining issues; I'll take an eye - we're not in a hurry :)

gagnieray:
> I saw a problem about autocomplete, but on members list - but that was during development, I just took a quick look.

You mean on the member add form ?
At the moment, it's the only place where this error has been raised in the latest tests.
It will have to be addressed and skipped in the tests if possible, if it's not relevant.

Otherwise, the latest tests still show issues with contrasts :confused:
#833 should hopefully fix that.

trasher:
> > I saw a problem about autocomplete, but on members list - but that was during development, I just took a quick look.
> 
> You mean on the member add form ? 

Nope, on list. But seems I have different results on CI and locally.

On CI:
image

Local:
image

trasher:
THe following is not accessible:
```
<i class="ui user tie icon tooltip" aria-hidden="true"></i>
<span class="ui special popup">{{ _T("Staff member") }}</span>
```

It's OK when I remove the span. Did not find how to fix that.

trasher:
> It's OK when I remove the span. Did not find how to fix that.
Adding an `@aria-label` on the `<a>` did the trick; dunno if that's really correct though.

galette/templates/default/pages/status_list.html.twig
```diff
@@ -85,22 +88,25 @@
                         </td>
                         <td class="center actions_row collapsing">
                             <a
-                                    href="{{ url_for('editStatus', {'id': eid}) }}"
-                                    class="action single-edit"
+                                href="{{ url_for('editStatus', {'id': eid}) }}"
+                                class="action single-edit"
+                                aria-label="{{ _T("Edit '%s' field")|replace({'%s': entry.name}) }}"
```
trasher: This kind of hack worked here, but on other pages not.

I'm not satisfied with that; we need to duplicate text while it's not really needed. Links may have also a title, and text can be complex (conditions etc.). Doing like that we'll end with desynchronized text, an aria-labels this will not be visible.

trasher: So, the correct accessible version would be:
```
<i class="ui edit icon tooltip" aria-hidden="true"></i>
<span class="visually-hidden">{{ _T("Edit '%s' field")|replace({'%s': entry.name}) }}</span>
```

Of course, this breaks the popup.

In the past (before FUI), Galette had a popup event that relied on the content a of child element; to prevent duplicate contents in code.
I've tried to do the same with FUI; I did not find a solution for now.

If that's not possible, we can use a macro or a template include for those kind of link; which would produce the expected output for us - but I'm not sure repeating systematically the same text in span content, title and/or aria-label is a good idea; that would duplicate contents without reason (not sure it's better event for a11y to have the text several times).

gagnieray:
> Nope, on list. But seems I have different results on CI and locally.
> 
> On CI: 
> image
> 
> Local: 
> image

That's weired. On the member list, indeed, the autocomplete attribute is used on the filters.
But the error should also be raised on the contributions list then...

Anyway, this error is raised to be sure that we use appropriate values for this attribute, or that the value "off" is "genuinely necessary": https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#accessibility

So, when the attribute is added by FUI, I think we can assume that it's OK.
Otherwise, we should check the value on the fields where we set the attribute ourselves (if any).

galette/templates/default/pages/status_list.html.twig
gagnieray: > In the past (before FUI), Galette had a popup event that relied on the content a of child element; to prevent duplicate contents in code. I've tried to do the same with FUI; I did not find a solution for now.
> 
> If that's not possible, we can use a macro or a template include for those kind of link; which would produce the expected output for us - but I'm not sure repeating systematically the same text in span content, title and/or aria-label is a good idea; that would duplicate contents without reason (not sure it's better event for a11y to have the text several times).

I haven't looked in details yet, but the solution to implement to prevent unnecessary duplications will depend on the use case of the tooltip/popup. 

And typically, regarding the `aria-label` attribute, it depends on the element on which it is used. If it's used for links/buttons that only display an icon, it should be valid. However, if it's used to display a description, it shouldn't be valid : https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label

gagnieray: For "icon-only" links, I think the new popup initialization I added in `common.js` does the job properly. What do you think ?

galette/templates/default/components/form.html.twig
```diff
@@ -106,6 +106,7 @@
                         {% endif %}
                         {% if entry.field_id == 'email_adh' %}
                             {% set size = "30" %}
+                            {% set autocomplete = "new-password" %}
```
gagnieray: Here, I used the "new-password" value to prevent the browser from accidentally filling in an existing password or offering assistance in creating a secure password. This happens to me using Firefox, and it also occurs in other fields on this form that I didn't modify for the moment : 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#new-password

However I am not sure it is really appropriate 🤔 
What do you think ?

galette/templates/default/pages/status_list.html.twig
trasher: That's exactly something like that I had in mind; thanks! I'll take an eye :)

galette/templates/default/components/form.html.twig
trasher: Hum... That does not really sounds appropriate; but also I see nothing else than "off" that could be relevant.

To me, this is a false positive; a `off` value for the SMTP password will not cause issues with accessibility; this is **not possible** to have autocomplete.
Therefore I propose to keep the original code; and add an exception in tests for this one (sohould be possible as far as I know).

galette/templates/default/elements/scripts.html.twig
```diff
@@ -134,7 +134,7 @@
                 _compactMenu();
     {% endif %}
 
-    {% if autocomplete %}
+    {% if autocomplete_options %}
```
trasher: I get the following error:
```
[18-Apr-2026 20:57:31 UTC] PHP Fatal error:  Uncaught Twig\Error\RuntimeError: Variable "autocomplete_options" does not exist in "elements/scripts.html.twig" at line 137. in galette/templates/default/elements/scripts.html.twig:137
```

trasher: I've fixed with `is defined`.

galette/templates/default/components/form.html.twig
trasher: I don't know why I was looking at something else... ... :D
I'll think about that.

tests/e2e/specs/a11y.spec.ts
```diff
@@ -67,7 +67,7 @@ test.describe('Accessibility', () => {
     await listPage.goto();
     await listPage.memberTable.waitFor({ state: 'visible' });
 
-    const results = await axeBuilder(page).analyze();
+    const results = await axeBuilder(page).exclude('*[autocomplete="fomantic-search"]').analyze();
```
trasher, line 80: Not much we can do; it's fomantic internals:
```
node_modules/fomantic-ui/dist/semantic.js:                                .prop('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
node_modules/fomantic-ui/dist/semantic.js:                                .attr('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
node_modules/fomantic-ui/src/definitions/modules/dropdown.js:                                .prop('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
node_modules/fomantic-ui/src/definitions/modules/search.js:                                .attr('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
```

trasher:
A11y tests are green \\o//
Good job!

I saw you talked about not tested pages in one of your commits messages; it would be great to have a list of missing pages.
I guess I can have all possible routes from router, I'll take a look.

gagnieray:
> I saw you talked about not tested pages in one of your commits messages; it would be great to have a list of missing pages.

Here is a list of untested pages where I fixed things : 
- advanced-search ~(fix in my last commit about it is incomplete: changes in js are still required)~
- scheduled-payments
- scheduled-payment/{id}/add
- fields/core/configure
- payment-types
- import/model
- plugins
- public/members/list
- public/staff/list

galette/templates/default/elements/group.html.twig
```diff
@@ -89,7 +89,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnmanagers_small">
+                    <a href="#" class="jsonly ui labeled icon button" id="btnmanagers_small">
```
gagnieray:
```suggestion
                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnmanagers_small">
```
I think the "disabled" class should be kept for the moment because it actually prevents the user from clicking on the button (using the pointer-events CSS property) if javascript is not enabled in their browser.

In my opinion, removing this behaviour (which uses the additional "jsonly" class btw; see [`common.js`](https://github.com/galette/galette/blob/47eafffd5c5664efa6d32b539fe8eeac5d8cdaa9/ui/js/common.js#L221)) could be done later globally (as well as removing `<noscript>` elements) if you ultimately consider them useless, since Galette obviously cannot fully function without  javascript.

FYI, this precise behaviour (removing the "disabled" class if javascript is enabled) is also used on the advanced search page.

Regarding the contrast issue on disabled components, it is actually not an issue as stated in the [WCAG 2.1 Understanding Docs](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#intent) : 

> [...] Note that this requirement does not apply to _inactive_ user interface components.

Across the templates, the "disabled" class is only used on `<a>` elements in: 
- member show (to disable the "generate member card" button when relevant)
- preferences (to disable the "translate" buttons when relevant)
- paginations (to disable "next" or "previous" buttons when relevant)

It is also used correctly on forms elements (inputs, selects, buttons, ...) with the additional "disabled" attribute.

The only cases I could find where the "disabled" class may not be used properly are some [FUI labels](https://fomantic-ui.com/elements/label.html) or texts, but there are only a very few. I think only those cases should be reviewed for the moment to determine why this class has been applied to them. If it is only to lower their opacity, it should be done differently and fixed.

galette/templates/default/elements/group.html.twig
```diff
@@ -109,7 +109,7 @@
             <div class="active content field">
                 {% if group %}
                     {% if login.isAdmin() or login.isStaff() %}
-                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnusers_small">
+                    <a href="#" class="jsonly ui labeled icon button" id="btnusers_small">
```
gagnieray:
```suggestion
                    <a href="#" class="jsonly disabled ui labeled icon button" id="btnusers_small">
```

gagnieray (dismissed):

galette/templates/default/elements/group.html.twig
trasher: Well, the problem here is not really an incorrect contrast, but the disabled state has been removed from the dom, and when playwright test colors, it's (from an UI pov) somewhere between disabled and not (that's why logged color codes does not suit any existing color). 
Maybe is there a transition effect that could be dropped?

I have no fix for that except dumb waiting for arbitrary seconds... Tests are already long and complex enough, I'm trying not to add hacks. I guess ths will end with an exclusion on problematic elements.

This specific case should probably have been handled differently (on list array for example, select all/none are added from JS, nothing JS only related is present on the page.
A good solution would have been to propose a no JS way to add manager and members to groups. 

TBH; it's been years since someone complains about javascript mandatory support - I do not want to keep wasting time for that kind of considerations.

Regarding standards compliance, I guess the fact is the element is **not** disabled from a semantic pov, but just visually. There is maybe a specific attribute to set; but since this is not the problem on that page, we'll see that later.

trasher: I put back `disabled` class, and excluded the buttons - seems the best option for now.

trasher:
> Here is a list of untested pages where I fixed things :

Thanks!

>     * advanced-search ~(fix in my last commit about it is incomplete: changes in js are still required)~
>     * scheduled-payments
>     * fields/core/configure
>     * payment-types
>     * import/model
>     * plugins

All added :) I'll add some other later.

>     * scheduled-payment/{id}/add

This one is too complex case for now.

>     * public/members/list
>     * public/staff/list

By default, public pages are accessible for up-to-date members. I've not been able to change preferences from Playwright - I cannot test them as anonymous for the moment :'(

trasher:
This PR is quite huge now; I'll do some last checks, and I'll go ahead and merge.

Not all is tested; but it will be easy to add new tests on new pages :)
Also, I guess it's possible to factorize using Playwright capacities; but I'm a newbie :p

tests/e2e/fixtures/a11y.fixture.ts
```diff
@@ -36,7 +36,7 @@ import type { Page } from '@playwright/test';
  */
 export function axeBuilder(page: Page): AxeBuilder {
   //FIXME: real A11y issues must be fixed in order to activate all tags.
-  return new AxeBuilder({ page }).withTags(['wcag21a'/*, 'wcag21aa', 'RGAAv4'*/]);
+  return new AxeBuilder({ page }).withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa', 'RGAAv4', 'best-practice', 'EN-301-549']);
```
trasher, line 40:
```suggestion
  // See available tags: https://www.deque.com/axe/core-documentation/api-documentation/#axecore-tags
  return new AxeBuilder({ page }).withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa', 'RGAAv4', 'best-practice', 'EN-301-549']);
```

PR #836, opened 2026-04-19
Revive Db::tableExists with a better solution

trasher:
On plugins side, `Db::tableExists` was replaced because of poor perfs with something like:

```
try {
    $this->zdb->execute($this->zdb->select('table')->limit(1));
    return true;
} catch (\Throwable $e) {
    if (!$this->zdb->isMissingTableException($e)) {
        throw $e;
    }
}
return false;
```

This works; but this is quite complex for plugins. Also, this generates an error in the logs when table does not exists (thown from Db::execute.

My proposition is to rewrite DB::tableExists with new way; and hide error messages when needed.


trasher:
Manually merged

PR #839, opened 2026-04-22
Improve pages titles

trasher:


trasher:
@gagnieray I'd like your opinion if you have time :)

The goal is to harmonize titles ans make them more accessible (without being too complex). It should be changed on every page (translations must be updated, I tried to change as less as possible).

gagnieray (commented):
This new formatting of the page titles LGTM :+1:

galette/templates/default/pages/contributions_types_list.html.twig
```diff
@@ -161,7 +161,7 @@
             " %}
             {% include "elements/js/modal_action.js.twig" with {
                 selector: ".single-edit",
-                modal_title_twig: _T("Edit contribution type")|e("js"),
+                modal_title_twig: (_T("Contribution type") ~ ' - %1$s')|e("js"),
```
trasher: @gagnieray could you please help on this one?

I'd like thee modal to have the same title than the directed accessed page; but I did not find how to replace the variable in the modal title :/

gagnieray: I see no other way to do that than to modify/extend `modal_action.js.twig` and use JS for this case.

trasher: Great, thanks!

PR #840, opened 2026-04-22
Fix langs with regions

trasher:
closes #2008

Before (see duplicated "Português"):
Capture d’écran du 2026-04-23 00-25-26

After:
Capture d’écran du 2026-04-23 00-42-33

No longer duplicates, the region (Brasil) is displayed. I made exceptions:
- `US` would have been displayed as "English, United States",
- When the region and locale code are the same (`fr_FR`, `de_DE`, `it_IT`, ...)

No idea if I should add an exception for `nb_NO`.

### DIFF
--- a/galette/templates/default/elements/language.html.twig
+++ b/galette/templates/default/elements/language.html.twig
@@ -35,19 +35,22 @@
     <div class="image header title"{% if ui == 'item'%} data-fold="fold-language" tabindex="0"{% endif %}>
 {% endif %}
         <i class="icon language" aria-hidden="true"></i>
-        <span>{{ galette_lang }}</span>
+        <span>{{ i18n.getName() }}</span>
         <i class="dropdown icon" aria-hidden="true"></i>
 {% if header == true %}
     </div>
 {% endif %}
     <div class="{{ content_classes }}">
-{% for langue in languages %}
-    {% if langue.getAbbrev() != galette_lang %}
-        <a href="?ui_pref_lang={{ langue.getID() }}"
-           title="{{ _T("Switch locale to '%locale'")|replace({"%locale": langue.getName()}) }}"
-           class="item"
+{% for glang in languages %}
+    {% if glang.getAbbrev() != galette_lang %}
+        {% set html_lang -%}
+            <span{% if glang.isRtl() %} dir="rtl"{% endif %} lang="{{ glang.getAbbrev() }}">{{ glang.getName() }}</span>
+        {%- endset %}
+        <a href="?ui_pref_lang={{ glang.getID() }}"
+           class="item tooltip"
+           data-html="{{ _T("Switch locale to '%locale'")|replace({"%locale": html_lang})|e }}"
         >
-            {{ langue.getName() }} <span>({{ langue.getAbbrev() }})</span>
+            {{ html_lang }} ({{ glang.getLongId(true) }})
         </a>
     {% endif %}
 {% endfor %}
--- a/galette/lib/Galette/Core/I18n.php
+++ b/galette/lib/Galette/Core/I18n.php
@@ -219,11 +219,13 @@ class I18n
     /**
      * Get long identifier
      *
+     * @param bool $noutf if true, remove utf part
+     *
      * @return string current language long identifier
      */
-    public function getLongID(): string
+    public function getLongID(bool $noutf = false): string
     {
-        return $this->longid;
+        return $noutf ? str_replace('.utf8', '', $this->longid) : $this->longid;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


trasher:
I've refined exclusions on `nb_NO`; according to https://www.localeplanet.com/icu/iso639.html

trasher:
After few adaptations; the final result is:
image

galette/templates/default/elements/language.html.twig
```diff
@@ -35,19 +35,19 @@
...
-        <a href="?ui_pref_lang={{ langue.getID() }}"
-           title="{{ _T("Switch locale to '%locale'")|replace({"%locale": langue.getName()}) }}"
+{% for glang in languages %}
+    {% if glang.getAbbrev() != galette_lang %}
+        <a href="?ui_pref_lang={{ glang.getID() }}"
+           title="{{ _T("Switch locale to '%locale'")|replace({"%locale": glang.getName()}) }}"
            class="item"
         >
-            {{ langue.getName() }} <span>({{ langue.getAbbrev() }})</span>
+            {{ glang.getName() }} <span lang="{{ glang.getAbbrev() }}">({{ glang.getLongId(true) }})</span>
```
gagnieray:
```suggestion
            {{ glang.getName() }}
```

As the language name is now used to show the active language instead of the language ID, is it still relevant to show the language ID in the list of other available languages ?

gagnieray:
> [...] I made exceptions:
> 
> * `US` would have been displayed as "English, United States",
> * When the region and locale code are the same (`fr_FR`, `de_DE`, `it_IT`, ...)
> 
> 
> No idea if I should add an exception for `nb_NO`.

When the region and locale code are the same, **and there isn't any other existing regional version in the list** it seems logical to make exceptions not to display the region, because it can obviously be considered as the "standard" version of the language.

But not in other cases, in my opinion.

Regarding the English language, it is a special case. I may be wrong, but internationally, the "standard" version is the UK one. Not the US one. So I think that it should normally be displayed as "English, United States". However, since it is the only available English version, I think it is acceptable to display it as "English" only.

Regarding the Norwegian languages, from what I understand, they should not be added as exceptions. It seems to be another special case where, unlike "English," each version has a different name. Therefore, each version should be displayed as "Language name, Region."

galette/templates/default/elements/language.html.twig
trasher: Yes, for me :D

gagnieray: Alright then! :grin:

galette/templates/default/elements/language.html.twig
```diff
@@ -35,19 +35,21 @@
...
-        <a href="?ui_pref_lang={{ langue.getID() }}"
-           title="{{ _T("Switch locale to '%locale'")|replace({"%locale": langue.getName()}) }}"
-           class="item"
+{% for glang in languages %}
+    {% if glang.getAbbrev() != galette_lang %}
+        <a href="?ui_pref_lang={{ glang.getID() }}"
+           class="item icon-only"
         >
-            {{ langue.getName() }} <span>({{ langue.getAbbrev() }})</span>
+            <span lang="{{ glang.getAbbrev() }}">{{ glang.getName() }}</span> ({{ glang.getLongId(true) }})
```
trasher: `@dir` is missing

gagnieray: > `@dir@ is missing

Sorry, but what does that mean ?

trasher: Each part of the text displayed in another language than the main one must declare its language, and its direction if it's not main page one ;)

trasher: (also, I fixed my original message, was mixing gh and redmine ^^)

trasher: I fogot to mention this is related to a11y only!

trasher:
> > [...] I made exceptions:
> > 
> > * `US` would have been displayed as "English, United States",
> > * When the region and locale code are the same (`fr_FR`, `de_DE`, `it_IT`, ...)
> > 
> > No idea if I should add an exception for `nb_NO`.
> 
> When the region and locale code are the same, **and there isn't any other existing regional version in the list** it seems logical to make exceptions not to display the region, because it can obviously be considered as the "standard" version of the language.
> 
> But not in other cases, in my opinion.

Globally, I do agree.

> 
> Regarding the English language, it is a special case. I may be wrong, but internationally, the "standard" version is the UK one. Not the US one. So I think that it should normally be displayed as "English, United States". However, since it is the only available English version, I think it is acceptable to display it as "English" only.

We'll keep as "English". Default Galette locale has always been `en_US`; and I won't accept en_GB alongside.

> Regarding the Norwegian languages, from what I understand, they should not be added as exceptions. It seems to be another special case where, unlike "English," each version has a different name. Therefore, each version should be displayed as "Language name, Region."

Yes, but in those cases; exceptions are done on the default version. As far as I understood `nb_NO` is the default variant just as `fr_FR`.
Since there is no alternative; this should not have any real impact for now.

TBH I also made like that because this was too long in display :]

galette/templates/default/elements/language.html.twig
gagnieray: Oh OK! Something like this is not possible ?: 
```suggestion
            <span{% if glang.isRtl() %} dir="rtl"{% endif %} lang="{{ glang.getAbbrev() }}">{{ glang.getName() }}</span> ({{ glang.getLongId(true) }})
```

trasher: It is just exactly what should be done, GG :wink:

ui/js/common.js
```diff
@@ -87,6 +88,16 @@ var _bindFomanticComponents = function() {
             addTouchEvents: false,
         })
     ;
+    $iconOnly.each(function() {
+        const $this = $(this);
+        $this.popup({
+            variation: 'inverted',
+            inline: false,
+            addTouchEvents: false,
+            html: $this.find('.visually-hidden').html(),
```
trasher: I had in mind using `html` here instead of `content` would render HTML correctly, but tags are displayed :/ 
I can't find any FUI example with a HTML content.

trasher: Found :)

PR #866, opened 2026-06-02
Fix tooltips issues

trasher:
closes #2013

### DIFF
--- a/galette/templates/default/pages/plugins.html.twig
+++ b/galette/templates/default/pages/plugins.html.twig
@@ -36,15 +36,16 @@
     {% endif %}
     <tr>
         <td data-scope="row">
-            <span class="ui icon"{% if not is_enabled %} data-tooltip="{{ disabled_cause }}"{% endif %}>
-                <i class="{% if is_enabled %}check circle green{% else %}exclamation circle orange{% endif %} icon"></i>
+    {% if is_enabled %}
+            <i class="ui check circle green icon" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


galette/templates/default/pages/plugins.html.twig
gagnieray: Here is another markup proposal for the tooltips on the plugins page. What do you think?

```Twig
            {% if is_enabled %}
                <i class="ui check circle green icon" aria-hidden="true"></i>
                <a href="{{ url_for(plugin.route ~ "Info", {"plugin": name}) }}" class="info-plugin">{{ plugin.name }} ({{ name }})</a>
            {% else %}
                <span class="icon-only">
                    <i class="ui exclamation circle orange icon" aria-hidden="true"></i>
                    <span class="visually-hidden">{{ disabled_cause|trim }}</span>
                </span>
                {{ plugin.name }} ({{ name }})
            {% endif %}
```

ui/js/common.js
gagnieray, line 1: Those changes should be reverted as well as commit a016977

gagnieray (changes requested):

galette/templates/default/pages/plugins.html.twig
trasher: I've applied proposal, thanks

ui/js/common.js
trasher: Done; let's now see if a11y tests are still green

trasher: Looks like OK, but as far as I rmember, the issue was not present everytime. Wait & See; if it reappears, we'll see how to fix.

gagnieray (approved):

PR #879, opened 2026-06-23
Fix begin of membership calculation when using pref_beg_membership

trasher:


trasher:
Manually merged on my side

PR #887, opened 2026-07-08
Batch mailings

trasher:
### Mailing queue

Mass mailings were sent as a **single message with every member in BCC**, in one
blocking request, with no rate control. This is incompatible with mail servers
that restrict the number of recipients per message or the sending rate
(hourly/daily quotas). This PR makes mass mailing compatible with restrictive
servers.

**Fully backward compatible**: with default preferences the
behaviour is unchanged (a single BCC message, sent synchronously).

### Cleanup — remove qmail
`METHOD_QMAIL` is obsolete (qmail unmaintained since ~2007) and redundant with `sendmail`. It is removed
from the code and the preferences UI; existing installs on that method are
migrated to `sendmail` during upgrade.


trasher:
Related doc: https://galette--50.org.readthedocs.build/en/50/usermanual/adherents.html#sending-a-large-mailing

PR #896, opened 2026-08-01
Improvements on the list of mailings

gagnieray:
While running tests to investigate an issue posted on the users' mailing list related to email signatures, I noticed that improvements could be made to the list of mailings in Galette :grin: 
  
Unsent emails cannot actually be duplicated to be used as a template to create a new email. Only sent emails can be duplicated in that way. Unsent emails can only be modified/edited, so they should be considered drafts.

That’s why I used different icons in the "Sent" and "Actions" column to make the distinction clearer. 

I also updated the visually hidden text associated with the icon in the "Actions" column accordingly, and I added the visually hidden text that was missing in the "Sent" column to improve accessibility.  

Still to improve accessibility, I corrected the use of abbreviations in the columns' headers.  

And finally, I added a legend to the table for better understanding.

mailings

### DIFF
--- a/galette/templates/default/pages/mailings_list.html.twig
+++ b/galette/templates/default/pages/mailings_list.html.twig
@@ -163,13 +163,17 @@
             <td data-col-label="{{ _T('Sender') }}">{% if log.mailing_sender == 0 %}{{ _T('Superadmin') }}{% else %}{{ log.mailing_sender_name }}{% endif %}</td>
             <td class="collapsing" data-col-label="{{ _T('Recipients') }}">{{ log.mailing_recipients|length }}</td>
             <td data-col-label="{{ _T('Subject') }}">{{ log.mailing_subject }}</td>
-            <td class="center collapsing" data-col-label="{{ _T('Attachments') }}">{{ log.attachments }}</td>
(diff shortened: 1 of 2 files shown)


galette/templates/default/pages/mailings_list.html.twig
```diff
@@ -200,7 +211,65 @@
     {% endfor %}
 {% endblock %}
 
+{% block legend %}
+    <div id="legende" title="{{ _T("Legend") }}" class="ui modal">
```
trasher, line 215: I'm not sure a legend is really required here; it seems it almost repeats controls on page.

trasher (commented):
I'll have to test (no time for now) - thanks for the PR :)

galette/templates/default/pages/mailings_list.html.twig
gagnieray: I think it can be useful on mobile because tooltips can't be displayed in this case.

trasher: Indeed, I did not had that in mind

trasher (approved):

PR #908, opened 2026-08-10
Switch to HTTP headers for CSRF checks

trasher:


trasher:
ping @gagnieray :)

This change affects plugins: `components/forms/csrf.html.twig` file has been removed, so this will be a fatal error at display.
I should have kept an empty file; but I guess it's better to have correct fix - next Galette release already have BC breaks anyway.

Is that OK with you?

galette/templates/default/pages/403.html.twig
```diff
@@ -0,0 +1,35 @@
...
+        {% include 'elements/header.html.twig' %}
+    </head>
+    <body class="{% if preferences.pref_enable_custom_colors %}custom-colors{% endif %}">
+        <div class="ui container">
+            <header class="ui basic center aligned segment">
+                <img src="{{ url_for('logo') }}" width="{{ logo.getOptimalWidth() }}" height="{{ logo.getOptimalHeight() }}" alt="[ Galette ]" />
+            </header>
+            <main>
+                <div class="ui red message">
+                    <h1 class="ui center aligned header">{{ _T("Unauthorized") }}</h1>
```
gagnieray:
```suggestion
                    <h1 class="ui center aligned header">{{ _T("Access denied") }}</h1>
```

galette/templates/default/pages/403.html.twig
```diff
@@ -0,0 +1,35 @@
...
+                </div>
+                <div class="ui basic fitted center aligned segment">
+                    <p class="ui large text">{{ _T("You do not have permission for requested URL.") }}</p>
```
gagnieray:
```suggestion
                    <p class="ui large text">{{ _T("You don’t have permission to access this resource.") }}</p>
```

galette/lib/Galette/Renderers/Html.php
```diff
@@ -50,6 +51,9 @@ public function __invoke(Throwable $exception, bool $displayErrorDetails): strin
         if ($exception instanceof HttpNotFoundException) {
             $code = 404;
             $title = __('Page not found');
+        } elseif ($exception instanceof HttpForbiddenException) {
+            $code = 403;
+            $title = __('Unauthorized');
```
gagnieray:
```suggestion
            $title = __('Access denied');
```

galette/lib/Galette/Middleware/Csrf.php
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * Can the request be trusted, from a CSRF point of view?
+     *
+     * @param Request $request PSR7 request
+     */
+    private function isSafeFromCsrf(Request $request): bool
+    {
+        //`Sec-Fetch-Site` tells where the request comes from. It is a forbidden header
+        //name; browsers always send a legitimate value, scripts cannot alter it.
+        $fetch_site = $request->getHeaderLine('Sec-Fetch-Site');
```
gagnieray, line 93: After reading [OWASP CSRF prevention cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#fetch-metadata-headers), I suppose that only checking the `Sec-Fetch-Site` header is necessary, and that there’s no point checking the additional `Sec-fetch-*` headers?

gagnieray (commented):
> This change affects plugins: `components/forms/csrf.html.twig` file has been removed, so this will be a fatal error at display. I should have kept an empty file; but I guess it's better to have correct fix - next Galette release already have BC breaks anyway.
> 
> Is that OK with you?

Yes! 👍

galette/lib/Galette/Middleware/Csrf.php
trasher: Yep, other headers are optional and would not be used in Galette anyway :)

PR #920, opened 2026-08-21
Switch History methods and properties to protected

gagnieray:
Required for galette-plugins/plugin-stripe#58 and galette-plugins/plugin-helloasso#36


trasher (changes requested):
Tests are failing, more adaptations are needed

trasher (approved):

PR #922, opened 2026-08-22
A11y improvements, enhance Fomantic dropdown

trasher:


trasher:
Issue detected (and fixed) with IA agent

gagnieray:
I’ll take a look at these changes as soon as I have a moment :wink: 

But after a quick initial review, I’m not sure this is the right way to go. Otherwise, it would only be a "temporary" fix/hack, and it might be preferable to contribute to FUI to fix these issues "at the source" :thinking: 

In the issue regarding the accessibility of FUI dropdowns (fomantic/Fomantic-UI#1332), using templates is recommended. Do you know if the AI agent took this into account?

I don’t know whether this would help resolve all the issues that have been identified, but if it does, I think it would be a better solution to adopt. It will also probably minimize problems if a future FUI update introduces breaking changes to the dropdowns.

- https://fomantic-ui.com/modules/search.html#templates
- https://fomantic-ui.com/modules/dropdown.html#templates

trasher:
> I’ll take a look at these changes as soon as I have a moment 😉

Thanks!

> But after a quick initial review, I’m not sure this is the right way to go. Otherwise, it would only be a "temporary" fix/hack, and it might be preferable to contribute to FUI to fix these issues "at the source" 🤔
> 
> In the issue regarding the accessibility of FUI dropdowns ([fomantic/Fomantic-UI#1332](https://github.com/fomantic/Fomantic-UI/issues/1332)), using templates is recommended. Do you know if the AI agent took this into account?

I was not aware of templates, no idea if agent was. I'll ask it ;)

Regarding the original issue... It's 6+ years old :/ 
Work that has been done had only Galette in mind, it may be far from generic.

trasher:
> > In the issue regarding the accessibility of FUI dropdowns ([fomantic/Fomantic-UI#1332](https://github.com/fomantic/Fomantic-UI/issues/1332)), using templates is recommended. Do you know if the AI agent took this into account?
> 
> I was not aware of templates, no idea if agent was. I'll ask it ;)

Seems not adapted, for several reasons (see below). 

As the PR changes the DOM, that would be refused upstream - so only a part can be split. I'll update the PR with that in mind - this should resolve to a patch that may be proposed upstream (if you want to), and another part that would stay Galette specific.

About upstream on that point, I'm not sure we really should expect any update... Their current "dev" version does not seem to include any change on that point - even if you propose a change that is accepted, it can take a while to reach Galette :/

Full agent answer about templates:

I did look at the templates after your pointer, and I don't think they can carry this — here's the detail.

`templates.dropdown` / `templates.menu` only build the **inner** HTML of a dropdown, once, at construction time. Three things fall outside of them:

- **The container.** `templates.dropdown` returns the caret, the text div and the menu. The `.ui.dropdown` element itself is built four lines above the template call, by `$('<div />').attr('class', $input.attr('class'))` (`src/definitions/modules/dropdown.js` L415 in 2.9.4, template called at L419). That element is where `role="combobox"`, `aria-expanded`, `aria-controls` and the accessible name have to go.
- **The name.** `templates.dropdown(select, …)` receives the output of `get.selectValues()` (L2090): `{placeholder, values: [{name, value, text, disabled}]}`. The `id` and the `name` of the `<select>` are not in there, so there is no way to reach the `<label for="…">` — which is the whole bug, since that label names a `<select>` the component then hides with `display: none`.
- **The state and the keyboard.** `aria-expanded`, `aria-selected` and `aria-activedescendant` change on every open and every cursor move; templates run once. And Enter/Space/Escape/Home/End, plus the focus the component drops on `<body>` after a pick, are behaviour, not markup.

Templates also never run for the dropdowns whose markup we write ourselves — 13 of those in our templates, against 64 `<select class="ui … dropdown">`. Those 13 are the navigation menus, which is exactly the disclosure half of this PR.

On robustness across FUI upgrades I think it actually cuts the other way: overriding a template means vendoring ~150 lines of upstream code, and the signature has already changed between 2.9.4 (`dropdown(select, fields, preserveHTML, className)`) and the 2.10 branch (`dropdown(select, settings)`). The current PR only reads the `active` / `visible` / `selected` classes off the DOM, which is the most stable part of the component's contract — unchanged since Semantic UI.

On contributing upstream, you're right in principle and I'd rather do that too. Where it stands today: the `develop` branch has zero `role=` and a single ARIA attribute in the whole dropdown module — the same `aria-labelledby` on the search input as 2.9.4. #1332 is six years old and tagged `help-wanted`; lubber-de says there himself that a proper fix needs DOM-structure changes and is "not desired by many users of SUI/FUI", and Jookia reports the same conclusion from the Gitea side.

So here's a middle ground I'd like to try: **split the work in two.**

1. Everything that does not change the DOM structure — roles, `aria-expanded` / `aria-selected` / `aria-activedescendant`, the name taken from `label[for]`, and the missing keys — moves *inside* `dropdown.js`, where the module already has the element and its own show/hide/select hooks (no `MutationObserver` needed). That becomes a patch applied to `semantic/src/` at build time, the same way we already patch TCPDF from `composer.json`, and the very same diff is what I'd propose upstream.
2. Only the disclosure pattern for menu dropdowns stays in Galette's own JS, because it inserts a wrapper element to avoid `nested-interactive` (the menu being a child of its own trigger) — that's the part upstream would legitimately push back on.

That way, if #1332 ever moves, we drop the patch and keep almost nothing of our own.

trasher:
> As the PR changes the DOM, that would be refused upstream - so only a part can be split. I'll update the PR with that in mind - this should resolve to a patch that may be proposed upstream (if you want to), and another part that would stay Galette specific.

I've opened a stacked PR (I often forget that does exist :D) with changes moved to an upstream patch.

ui/js/dropdown-a11y.js
```diff
@@ -0,0 +1,425 @@
...
+            //not initialised yet, or nothing to open
+            return;
+        }
+
+        //what the value is kept in, the two places the component looks at;
+        //"selection" is how it marks a dropdown standing for a form control,
+        //and some of them are built without any value element at all
+        var $value = $dd.children('select, input[type="hidden"]').first();
+        if ($value.length || $dd.hasClass('selection')) {
+            _asListbox($dd, $menu, $value);
```
gagnieray: `_asLisbox()` handles "simple" `.ui.drowpdown.selection` dropdowns (simple selection and search dropdowns).

Descriptions read with a screenreaders as well as the keyboard navigation work properly.

But it is incomplete and does not make [clearable selection drowpdowns](https://fomantic-ui.com/modules/dropdown.html#clearable-selection) fully accessible, nor the "paginated search dropdown" (the clearing and pagination buttons cannot be used with the keyboard). 

As a reminder, this last component does not exist in FUI; it is implemented in Galette’s Twig templates for its [HTML structure](https://github.com/galette/galette/blob/develop/galette/templates/default/components/forms/member_dropdown.html.twig), together with the corresponding [customized javascript code](https://github.com/galette/galette/blob/develop/galette/templates/default/components/forms/member_dropdown.html.twig).

ui/js/dropdown-a11y.js
```diff
@@ -0,0 +1,425 @@
...
+        } else {
+            _asDisclosure($dd, $menu);
```
gagnieray: `_asDisclosure()` handles all the others dropdowns and doesn't work properly. 

Here the AI agent only considered "menu" or "button" dropdowns. But there are many other variants (or variations) in FUI that could be impacted by this function.

Descriptions are not correctly read with a screenreader (I used [Orca](https://orca.gnome.org/); I can't test with other screenreaders as I don't have a Windows or MacOS machine). 

And I would say that keyboard navigation is "unpredictable" 😕

gagnieray (commented):
Well... To say the least, I think it’s going to be very difficult to come up with a "global" solution for making dropdowns accessible, even with the help of an AI agent 🤯 

I let you read through my initial comments. 

The next ones will be in French. I think it’ll be easier to provide clear explanations that way. I might also need to prepare some screencasts to illustrate things 😜

trasher:
I put back this one to draft for now. It appears to be far more complex than I initially think.

> Well... To say the least, I think it’s going to be very difficult to come up with a "global" solution for making dropdowns accessible, even with the help of an AI agent 🤯

Looks like, yes :(

> The next ones will be in French. I think it’ll be easier to provide clear explanations that way. I might also need to prepare some screencasts to illustrate things 😜

:+1:

trasher:
> I put back this one to draft for now. It appears to be far more complex than I initially think.
> 
> > Well... To say the least, I think it’s going to be very difficult to come up with a "global" solution for making dropdowns accessible, even with the help of an AI agent 🤯
> 
> Looks like, yes :(

Just a precision: the goal was to improve, if we can end with a working/satisfying solution - I'll already "be happy".
Maybe a simplier solution that just fixes major issues would be an option. I'll have to take an eye.

gagnieray:
Voici comment ça se comporte avec le lecteur d'écran [Orca](https://orca.gnome.org/) (active le son sur les vidéos, il est désactivé par défaut) : 

1. **les dropdowns "basiques"** (`selection` et `search`) basés ou non sur une balise `select` : ça fonctionne donc a priori correctement : 
   
   https://github.com/user-attachments/assets/ba5a53cc-865a-4880-83e7-2dc9199e38c7

2. **le menu déroulant de sélection de la langue** : les descriptions ne fonctionnent pas, et la navigation au clavier est "doublée". On peut naviguer avec les flèches du clavier, mais aussi avec la touche tabulation car ce sont des liens dans la liste déroulante. Et alors "visuellement" c'est le bordel :sweat_smile: 
   
   https://github.com/user-attachments/assets/4f1f6bc0-8b90-42f3-9be3-e806c0215dbf

3. **le menu déroulant des actions groupées** : on peut commencer la navigation au clavier avec les flèches, mais les descriptions lues ne correspondent pas à l'entrée sélectionnée dans le menu. Puis on peut avec une tabluation supplémentaire accéder à une autre navigation au clavier pour laquelle le composant est décrit, mais pas ses entrées :
   
   https://github.com/user-attachments/assets/cee133f6-7187-4234-8ec2-39b0013f96d4

4. **le composant de recherche d'un membre** : comme pour les dropdowns `selection` et `search` les descriptions et la navigation dans les entrées fonctionnent, mais impossible d'utiliser la pagination ou d'effacer une sélection : 
   
   https://github.com/user-attachments/assets/e1bcd756-d7de-4df2-9bed-28fc1cc62f59

trasher:
Ce test, c'est avec la PR ? Du coup... c'est moins pire à l'origine ?

gagnieray:
> Ce test, c'est avec la PR ? 

Oui.


> Du coup... c'est moins pire à l'origine ?

Oui et non :sweat_smile:  

Dans les cas n°1 et 4 c'est bien mieux avec la PR.

Pour les cas n°2 et 3 il y a à boire et à manger.
Selon moi c'est pire en l'état avec la PR car la navigation au clavier est "doublée" et la lecture des descriptions est aléatoire ou incorrecte.

Mais si ces problèmes peuvent être résolus pour chacun de ces cas précisément, ce sera forcément mieux :grin:

trasher:
2-3 choses ont été corrigés côté HTML, et la surcharge dropdown est plus légère.

Et au pire, on verra plus tard pour les dropdown. J'ai très clairement sous estimé l'ampleur de ce chantier (pour moi, ça aurait du prendre 2 à 3h maxi - c'est explosé déjà) - on récupérerait les modifs HTML déjà.

Dans certains cas particuliers (langue par exemple), il serait éventuellement possible de mettre l'actuel en aria-hidden et proposer un menu simple et accessible à côté. Ce n'est pas une solution qui m'enchante - mais ce serait sûrement la plus accessible (c'est sûrement une idée débile).

Mais du coup, je comprends mieux pourquoi l'issue d'origine est ancienne et n'avance pas :/
Si upstream est incapable de corriger, ou même d'améliorer quelque peu ce composant - je doute que j'y arriverai.

gagnieray:
> Et au pire, on verra plus tard pour les dropdown. J'ai très clairement sous estimé l'ampleur de ce chantier (pour moi, ça aurait du prendre 2 à 3h maxi - c'est explosé déjà) - on récupérerait les modifs HTML déjà.

Si il ne fallait garder qu'une chose du code javascript de cette PR, ce serait uniquement ce qui permet la description des dropdowns `selection` et `search` dans tous les formulaires.

La tentative pour rendre accessible le bouton d'effacement d'une sélection et les boutons de pagination est belle, mais elle est inefficace :confused: 

Ça a des effets de bord sur la navigation au clavier pour le bouton d'effacement.
Et ça n'a strictement aucun effet sur les boutons de pagination qui restent totalement inaccessibles.


> Dans certains cas particuliers (langue par exemple), il serait éventuellement possible de mettre l'actuel en aria-hidden et proposer un menu simple et accessible à côté. Ce n'est pas une solution qui m'enchante - mais ce serait sûrement la plus accessible (c'est sûrement une idée débile).

Pour les dropdowns utilisés dans la navigation, je ne pense pas que ce soit la solution adequate. 

Car il n'y a pas que le menu de la langue qui est concerné. À mon avis une solution plus "générique" pourrait être trouvée en changeant la "variante" des dropdowns utilisées dans cette situation. Mais plus globalement je pense qu'il faudrait commencer par corriger correctement les défauts d'accessibilité de la navigation dans les templates dans son ensemble.


> Mais du coup, je comprends mieux pourquoi l'issue d'origine est ancienne et n'avance pas :/ Si upstream est incapable de corriger, ou même d'améliorer quelque peu ce composant - je doute que j'y arriverai.

Oui je pense la même chose. J'en arrive même à me demander si il ne faudrait pas envisager de changer de framework pour l'UI car le composant des dropdowns n'est a priori pas le seul à poser des problèmes d'accessibilité :sweat_smile:

trasher:
> Si il ne fallait garder qu'une chose du code javascript de cette PR, ce serait uniquement ce qui permet la description des dropdowns `selection` et `search` dans tous les formulaires.

Ça marche.

> Mais plus globalement je pense qu'il faudrait commencer par corriger correctement les défauts d'accessibilité de la navigation dans les templates dans son ensemble.

Oui tout à fait - j'étais parti du principe que les dropdowns sont utilisées un peu partout.

> Oui je pense la même chose. J'en arrive même à me demander si il ne faudrait pas envisager de changer de framework pour l'UI car le composant des dropdowns n'est a priori pas le seul à poser des problèmes d'accessibilité 😅

Mouarf... Ce serait un travail conséquent :/
Et pour viser quoi ? Du bootstrap standard ?

trasher:
> > Oui je pense la même chose. J'en arrive même à me demander si il ne faudrait pas envisager de changer de framework pour l'UI car le composant des dropdowns n'est a priori pas le seul à poser des problèmes d'accessibilité 😅
> 
> Mouarf... Ce serait un travail conséquent :/ Et pour viser quoi ? Du bootstrap standard ?

Et à vrai dire, même pas sûr que ça change vraiment quoi que ce soit. Je ne suis clairement pas partant pour changer de nouveau de framework (il faudrait vraiment des arguments en béton :D).

gagnieray:
> > J'en arrive même à me demander si il ne faudrait pas envisager de changer de framework pour l'UI car le composant des dropdowns n'est a priori pas le seul à poser des problèmes d'accessibilité 😅
> 
> Mouarf... Ce serait un travail conséquent :/ 

C'est certain. Mais vouloir que Galette soit accessible dans le plus de cas de figure possible s'avère déjà être un travail énorme en fin de compte.

Corriger un à un les problèmes relevés par les tests automatisés a11y, en soi, ce n'est pas toujours très compliqué. Mais dès que ça touche à des composants d'interface, je me rends compte qu'en conditions réelles ce n'est absolument pas simple.

Par exemple, je viens de tester sur Android avec le lecteur d'écran [Talkback FOSS](https://talkbackfoss.app/), et ça n'a rien à voir avec Orca. Donc je me dis que même sur ordi, selon le lecteur d'écran, il doit y avoir des comportements différents, et ce que l'on pourrait penser être réglé ne l'est pas forcément :dizzy_face:  

> Et pour viser quoi ? Du bootstrap standard ?

Dans l'immédiat je ne me suis pas posé la question. En fait je ne crois pas qu'il existe un framework réellement adapté à toutes les situations. Et peut-être qu'en fin de compte si on veut bien faire les choses il serait même plus simple de ne pas utiliser de framework du tout et de réimplémenter les choses spécifiquement pour Galette :pensive: 

C'était juste une réflexion que je me suis fait. Pour le moment, continuer de faire au mieux avec FUI ne sera jamais perdu et aidera à prendre la bonne décision si un jour la question de l'abandonner s'impose :wink:

ui/js/dropdown-a11y.js
```diff
@@ -0,0 +1,365 @@
...
+            $control.attr('aria-required', 'true');
+        }
+        if ($value.prop('disabled') || $dd.hasClass('disabled')) {
+            $control.attr('aria-disabled', 'true');
+        }
+
+        $menu.attr('role', 'listbox');
+        if ($value.prop('multiple')) {
+            $menu.attr('aria-multiselectable', 'true');
+        }
```
gagnieray, line 124:
```suggestion
```
Je n'ai pas souvenir que des dropdowns à sélection multiple soient utilisés. Si tel est bien le cas, ces lignes ne sont pas nécessaires.

ui/js/dropdown-a11y.js
```diff
@@ -0,0 +1,365 @@
...
+    var _setup = function($dd) {
+        if ($dd.data('a11yBound')) {
+            return;
+        }
+
+        //a submenu, inside the menu of another dropdown: it belongs to what
+        //is around it, and describing it on its own would contradict that
+        if ($dd.parentsUntil('body', '.ui.dropdown').length) {
+            return;
+        }
```
gagnieray:
```suggestion
```
Ça ne peut concerner que des dropdowns pour la navigation a priori, pas ceux utilisés dans les formulaires.

gagnieray (changes requested):

ui/js/dropdown-a11y.js
trasher: dans la recherche avancée on utilise des dropdown multiples

ui/js/dropdown-a11y.js
trasher: En effet, c'est du code mort - c'est supprimé :)

PR #925, opened 2026-08-23
Work on language selection

trasher:
Use lang full ID web formatted globally
Display current lang marked active in langs list
Drop `galette_lang` Twig variable (BC on some plugins, I'll submit PRs).

Can be reviewed per commit.


trasher:
@gagnieray I asked you a review for the language selector display mainly; but I'll merge this PR quickly  so I can move forward ;)
Pontential fixes can be done in a following PR.

gagnieray:
> Pontential fixes can be done in a following PR.

Yes. At first sight, there will be some changes to make. For example, I don’t see why adding CSS with `!important` rules only when the `active` class already exists in the template. I need to test "in situ". If necessary, I’ll open a PR :wink:

PR #929, opened 2026-08-24
Fix language menu

gagnieray:


galette/templates/default/elements/language.html.twig
```diff
@@ -22,7 +22,7 @@
 {% endif %}
         <i class="icon language" aria-hidden="true"></i>
         <span class="visually-hidden">{{ _T("Choose your language") }}</span>
-        <span>{{ i18n.getName() }} ({{ i18n.getLongId(true) }})</span>
+        <span>{{ i18n.getName() }}</span>
```
gagnieray, line 25: I don't think showing the language long ID is really useful 🤔

galette/webroot/installer.php
```diff
@@ -219,14 +219,44 @@
...
                     <i class="icon dropdown" aria-hidden="true"></i>
                     <div class="menu">
 <?php
 foreach ($i18n->getList() as $langue) {
+    if ($langue->getID() === $i18n->getID()) {
     ?>
-                        <a href="?ui_pref_lang=<?php echo $langue->getID(); ?>" lang="<?php echo $langue->getWebID(); ?>" class="item"><?php echo $langue->getName(); ?> <span>(<?php echo $langue->getID(); ?>)</span></a>
+                        <a href="?ui_pref_lang=<?php echo $langue->getID(); ?>"
+                           class="item tooltip"
+                           data-html="<?php echo sprintf(_T('Current locale \'%1$s\''), $langue->getName()); ?>"
```
gagnieray, line 232: I couldn't find a way to use the exact original string that is already used in twig templates 😕

galette/webroot/installer.php
```diff
@@ -219,14 +219,48 @@
...
+                            >
+                                    <?php echo $langue->getName(); ?>
+                            </span>
+                        </a>
+        <?php
+    } else {
+        ?>
+                        <a href="?ui_pref_lang=<?php echo $langue->getID(); ?>"
+                           class="item tooltip"
+                           data-html="<?php echo sprintf(_T('Switch locale to \'%1$s\''), $langue->getName()); ?>"
```
gagnieray, line 250: Here neither, I couldn't find a way to use the exact original string that is already used in twig templates...

Should I simply remove the tooltips on the installer ? 🤔

galette/templates/default/elements/language.html.twig
trasher: No problem to remove it; since this is consistent everywhere ;)

galette/webroot/installer.php
trasher: Huh? This is the same string (make extract confirms).

galette/webroot/installer.php
trasher: You cannot use original `%locale` with sprintf. I replaced with` str_replace` on this one (I've already fixed several similar cases - but not yet this one.

trasher (approved):

galette/webroot/installer.php
trasher: I merge this one, if something must be fixed in translations - I'll do.
Anyway, translations won't be extracted for now; there are new strings that I'd like to get rid of before.

gagnieray: I must have been tired when I wrote that or confused it with the case that followed using `%locale`. Indeed they are the same. The only differences from the original are just escape characters :sweat_smile:

PR #931, opened 2026-08-25
Report a missing asset instead of a blank error

trasher:
I'm frequently switching branches while working, that easily leaves Galette without its default logo - leaving a blank 500 page in the browser and an exception in logs.

Now returns a self-contained page instead.


trasher:
Merged

PR #933, opened 2026-08-26
Replace links with buttons that will properly act

trasher:
I'm not sure for the new CSS, but I did not find correct existing solution.


gagnieray:
> I'm not sure for the new CSS, but I did not find correct existing solution.

OK. Using buttons requires further adjustments to the CSS. I’ll look into it ASAP :wink:

trasher:
OK, thanks a lot.
I'll merge this stack; we'll make corrections in new PRs ;)

PR #935, opened 2026-08-26
Few more a11y fixes

trasher:
Te be reviewed per commit


trasher:
This will be the last of the batch; I'll work on other subjects than a11y ;)

gagnieray (approved):

PR #945, opened 2026-08-28
Dynamic fields of type file are not repeatable; closes #2025

trasher:


gagnieray (approved):
I did the same in #944 😆

PR #946, opened 2026-08-29
A11y improvements, enhance autosubmit dropdowns

gagnieray:


ui/js/common.js
```diff
@@ -158,21 +158,6 @@ var _keyboardNavigation = function() {
     }
 }
 
-/* Required for keyboard accessibility on simple dropdowns with autosubmit.
- */
-var _bindDropdownsAutosubmit = function() {
```
trasher, line 163: This one is still referenced in `galette/templates/default/pages/transaction_form.html.twig`

ui/js/dropdown-a11y.js
```diff
@@ -339,12 +345,119 @@ var _dropdownA11y = (function() {
...
+        );
+
+        if (!description) {
+            return;
+        }
+
+        //_setup() normally stores the control in a11yControl.
+        //The fallback is necessary when the dropdown is initialised
+        //asynchronously by Fomantic.
+        var $control = $dd.data('a11yControl');
```
trasher: Already defined at the beginning of the function

trasher:
That seem globally correct to me; I asked agent for a review, and it found a possible logic duplication issue - I'm not sure it is entirely right - even if some remarks seems relevant. Here is the review output:

#### `_bindAutosubmit()` duplicates `_sweep()`, and the observer only runs one of them

`_sweep()` (`ui/js/dropdown-a11y.js:208-219`) already runs `_setup` + `_describeAutosubmit` + `_refreshState` on autosubmit dropdowns. `install()` then calls `_bindAutosubmit()`, which runs those exact same three calls again (lines 447-449) — and calls `_setup($dd)` twice in a row itself (427 and 447); the first one is a no-op, since nothing changes between the two except the Fomantic init that comes after it.

The duplication is mostly harmless, but the asymmetry it creates is not: the `MutationObserver` only calls `_sweep()`. So an autosubmit dropdown inserted after page load gets *described* but never *bound* — its autosubmit doesn't work. That's exactly the case of the infoline dropdown in `elements/list.html.twig`, which lives inside the `#main-container` that ajax mode replaces.

Folding the binding into `_sweep()` fixes the duplication and the dynamic case at once:

```js
var _sweep = function(root) {
    $(root || document).find('.ui.dropdown').each(function() {
        var $dd = $(this);
        if ($dd.hasClass('autosubmit')) {
            _bindAutosubmit($dd);   // keeps the data('autosubmitBound') guard
        }
        _setup($dd);
        if ($dd.hasClass('autosubmit')) {
            _describeAutosubmit($dd);
        }
        _refreshState($dd);
    });
};
```

`_bindAutosubmit` then takes a single `$dd` instead of a root, `install()` goes back to `_sweep(); _observe(); _keyboard();` and `refresh()` stays `_sweep(root)`.

One ordering note while you're in there: `_bindFomanticComponents()` deliberately excludes `.autosubmit` from its generic init (`ui/js/common.js:14`), so those dropdowns are only initialised by the autosubmit binding. With `install()` calling `_sweep()` first, `_setup()` now marks them `a11yBound` *before* Fomantic has initialised them, whereas `_bindDropdownsAutosubmit()` used to run first. It happens to work because the `.menu` and the hidden input are both in the markup, but binding before describing (as in the snippet above) keeps the previous, safer order.

gagnieray:
> That seem globally correct to me; I asked agent for a review, and it found a possible logic duplication issue - I'm not sure it is entirely right - even if some remarks seems relevant. Here is the review output: [...]

The logic duplication and the fact that an autosubmit dropdown inserted after page load gets described but never bound are real I think. That could happen in `galette/templates/default/pages/transaction_form.html.twig` with the list loaded in a modal.

I'll try to fix that. Thanks!

trasher:
I've pushed 2 new commits that make changes that were not blockers for this PR. If it's OK with you, we'll merge :)

gagnieray:
LGTM. You can merge :+1:

trasher (approved):

PR #966, opened 2026-09-06
Rework required php dependencies

trasher:
Add tests checking modules check and php version
are iso with composer.json declarations.
Fileinfo is now required, removing fallback code


Domenge:
Checked !

PR #967, opened 2026-09-07
Test emailing settings, fix send mail test button

trasher:
The test button built its message from the stored preferences, so an administrator who had just typed a new SMTP host tested the old one, with nothing on the page saying so. It also stayed clickable while emailing was disabled: the refusal only came back from the server, once the address modal had been filled in.

A second button has been added to test connection without sending anything.

Both tests now post the mail tab and run on what it carries. PreferencesSchema marks the preferences the transport is built from, and the controller overrides a clone of the live instance with those alone - nothing is written, so a server can be tried before it is saved.

Also, improve some failure messages.

### DIFF
--- a/galette/templates/default/pages/preferences.html.twig
+++ b/galette/templates/default/pages/preferences.html.twig
@@ -722,14 +722,33 @@
                             </div>
                         </div>
                         <br/>
-                        <a
-                            href="{{ url_for('testEmail') }}"
+    {% set mail_off = pref.pref_mail_method == constant('Galette\\Core\\GaletteMail::METHOD_DISABLED') %}
+                        <button
+                            type="button"
+                            data-href="{{ url_for('testEmailConnection') }}"
+                            id="btnmailconnect"
+                            class="ui labeled icon button{% if mail_off %} disabled{% endif %}"
+                            {% if mail_off %}disabled="disabled" aria-disabled="true"{% endif %}
+                        >
+                            <i class="plug icon" aria-hidden="true"></i>
+                            {{ _T("Test connection") }}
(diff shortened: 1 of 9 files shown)


trasher:
I merge, if there are things to change it can be done in another PR

galette/templates/default/pages/preferences.html.twig
```diff
@@ -722,14 +722,33 @@
...
-                            {{ _T("Test email settings") }}
-                        </a>
+                            {{ _T("Send a test email") }}
+                        </button>
+                        <div id="mailtest_off" class="ui small yellow message{% if not mail_off %} displaynone{% endif %}">
+                            {{ _T("Emailing is disabled: pick a sending method to run a test.") }}
+                        </div>
+                        <div id="mailtest_hint" class="ui small blue message{% if mail_off %} displaynone{% endif %}">
+                            {{ _T("Tests run on the settings displayed above, even those not saved yet.") }}
+                        </div>
```
gagnieray, line 751: To make sense in this position, this message should only be displayed when the selected method is SMTP or Gmail and use the word "below" instead of "above".

Or, the buttons and the message should be displayed below all the additional settings that appear when those methods are selected.

I'd vote for the second solution.
What do you think ?

trasher: Indeed there is a "problem" here. I'd prefer the admonition to stay close to the buttons; so at the bottom seems a good idea. :+1: 
Also, even is this not not really reliable, it checks for function or binary presence for other methods than smtp/gmail - and it's more simple to implement UI side ;)

PR #977, opened 2026-09-12
Fix link in lost password email; closes #2033

trasher:


trasher:
This was due to an error on a recent fix.

